PR #64, opened 2022-07-06
feat(UI): Use MUI Autocomplete component for improved `strings` input UX

dsvanlani:
### Proposed Change
Previously the `strings` type for parameter input was just a text input that parsed a string array by splitting on commas.

Now we're using the [Autocomplete](https://mui.com/material-ui/react-autocomplete/) component to display multiple values as chips.  It also has an improvement by hitting enter to delimit values.

Consider the `File` source type:

https://user-images.githubusercontent.com/64915094/177568569-5c9058b1-19fa-4570-963c-125d23697bdb.mov

### Source Type Changes:

Some `strings` input had a default of `[""]` which looked strange in the UI, I adjusted it to consistently use `[]` as the default.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


andykellr:
I think we can use `disableClearable` to get rid of the `X` on the input which can overlap the chips.

Screen Shot 2022-07-06 at 10 42 07 AM
 
The only downside to this control is that appears to be impossible to edit an existing value. You need to create a new chip and delete the old one.

Also the dropdown duplicating the input is also a little strange but maybe not that big of a deal.

andykellr (approved):

PR #67, opened 2022-07-07
Added agent section to quick start

asenkowski:
### Proposed Change
Added agent section to main README.md quick start.

##### Checklist
- [x] Changes are tested
- [x] CI has passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,16 @@ curl -fsSlL https://github.com/observiq/bindplane-op/releases/latest/download/in
 curl -fsSlL https://github.com/observiq/bindplane-op/releases/latest/download/install-macos.sh | bash -s --
 ```
 
+## Agent
+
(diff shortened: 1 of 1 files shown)


andykellr (approved):
Looks good. Should we mention running `--help` to see options for install like `--platform`?

README.md
```diff
@@ -75,6 +75,27 @@ curl -fsSlL https://github.com/observiq/bindplane-op/releases/latest/download/in
...
+```bash
+bindplanectl install agent
+```
+
+The generated command can then be copied and run on the agent host. To tailor the install command for a specific operating system, use the `--platform` flag. Valid values include:
+
+ * `linux`
+ * `linux-arm64`
+ * `windows`
+ * `macos`
```
andykellr, line 91:
```suggestion
 * `macos-amd64`
```

`macos` is an alias for arm64

asenkowski: Good catch. Fixed with latest commit

jsirianni: I think we should call them out explicitly instead of using the aliases

```
linux-arm64
linux-amd64
linux-arm
darwin-arm64
darwin-amd64
windows-amd64
```

PR #138, opened 2022-07-21
feat: shell tab completion instructions

algchoo:
### Proposed Change

added instructions for setting up tab completion for bash/zsh on linux/macos

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


docs/install.md
```diff
@@ -183,6 +183,45 @@ Copy the output and run the command on the system your wish to install the colle
...
+
+To setup zsh completion for bindplanectl on MacOS:
+1. Include the following lines in `~/.zshrc`&nbsp;
+```
+autoload -Uz compinit
+compinit
+```
+2. Locate `fpath` by running `echo $fpath`, there may be several listed, some may not exist, use an existing one in the next step.
+3. Run the following command to generate the zsh tab completion script.\
+`bindplanectl completion --zsh ><YOUR FPATH HERE>/_bindplanectl`
```
jsirianni, line 222: zsh is a sub command, not a flag
```suggestion
`bindplanectl completion zsh ><YOUR FPATH HERE>/_bindplanectl`
```

jsirianni (changes requested):

dsvanlani:
I was not able to get completion working with these instructions on Mac and zsh.

jsirianni:
I was able to get this working on Linux (zsh) but not macOS. I think it is working generally, but we have some consistency issues between macOS installs.

docs/install.md
```diff
@@ -183,6 +183,45 @@ Copy the output and run the command on the system your wish to install the colle
 
 After running the collector install command, the collector will appear in the `bindplanectl get agents` output.
 
+### Shell Completion
+
+#### Linux: bash
+
+1. Verify that `bash-completion` is installed on the host
+2. `sudo su` Switching to root
+3. `bindplane completion bash >/etc/bash_completion.d/bindplane` appends the output to a file in the bash completion directory
```
jsirianni, line 192: We should use `sudo tee` when redirecting output to a file owned by root, when not running as root.

jsirianni, line 192: Lets always use bindplanectl when running commands, bindplane (binary) may not support cli in the future.

```suggestion
3. `bindplanectl completion bash >/etc/bash_completion.d/bindplanectl` appends the output to a file in the bash completion directory
```

docs/install.md
```diff
@@ -183,6 +183,45 @@ Copy the output and run the command on the system your wish to install the colle
...
+4. Restart the shell
+
+#### macOS: bash
```
jsirianni, line 195: I am okay with removing this for now
1) I cant get it to work
2) Most macOS users will be using zsh

jsirianni (changes requested):

docs/install.md
algchoo: removed this portion of the instructions

jsirianni (approved):

andykellr:
It would be nice to add an .oh-my-zsh plugin, like https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/gh/gh.plugin.zsh

andykellr:
I'm good with merging this, but this is what I did:

1. Verify that compinit is already in .zshrc

2.
```sh
compinit; echo $fpath # /Users/andy/.oh-my-zsh/completions in the list but doesn't exist
mkdir /Users/andy/.oh-my-zsh/completions
bindplanectl completion zsh >/Users/andy/.oh-my-zsh/completions/_bindplanectl
```

3. Restart zsh

dsvanlani (approved):

PR #159, opened 2022-07-27
feat: Users can duplicate a config via REST and UI

dsvanlani:
### Proposed Change
It can be useful to make a copy of an existing config if you wanted to adjust sources slightly.  

This PR:
- Adds a REST endpoint `POST /v1/configurations/:name/duplicate` will copy the `Spec` of a config into a new one, changing only the `metadata.name`, `metadata.id`, and `spec.selector.matchLabels` fields.
- Add a "Duplicate" Button on the configuration view page:
image
image

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/internal/rest/api.go
+++ b/internal/rest/api.go
@@ -48,7 +48,7 @@ func AddRestRoutes(router gin.IRouter, bindplane server.BindPlane) {
 	router.GET("/configurations", func(c *gin.Context) { configurations(c, bindplane) })
 	router.GET("/configurations/:name", func(c *gin.Context) { configuration(c, bindplane) })
 	router.DELETE("/configurations/:name", func(c *gin.Context) { deleteConfiguration(c, bindplane) })
-	router.PUT("/configurations/:name/duplicate", func(c *gin.Context) { duplicateConfig(c, bindplane) })
+	router.POST("/configurations/:name/duplicate", func(c *gin.Context) { duplicateConfig(c, bindplane) })
 
 	router.GET("/sources", func(c *gin.Context) { sources(c, bindplane) })
 	router.GET("/sources/:name", func(c *gin.Context) { source(c, bindplane) })
@@ -479,7 +479,7 @@ func deleteConfiguration(c *gin.Context, bindplane server.BindPlane) {
 
 // @Summary Duplicate an existing configuration
 // @Produce json
-// @Router /configurations/{name}/duplicate [put]
+// @Router /configurations/{name}/duplicate [post]
 // @Param 	name	path	string	true "the name of the configuration to duplicate"
 // @Param name	body	string	true "the desired name of the duplicate configuration"
 // @Success 201	"Successful Duplication, created"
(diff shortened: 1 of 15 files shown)


internal/rest/api.go
```diff
@@ -48,6 +48,7 @@ func AddRestRoutes(router gin.IRouter, bindplane server.BindPlane) {
 	router.GET("/configurations", func(c *gin.Context) { configurations(c, bindplane) })
 	router.GET("/configurations/:name", func(c *gin.Context) { configuration(c, bindplane) })
 	router.DELETE("/configurations/:name", func(c *gin.Context) { deleteConfiguration(c, bindplane) })
+	router.PUT("/configurations/:name/duplicate", func(c *gin.Context) { duplicateConfig(c, bindplane) })
```
andykellr, line 51: I think semantically a `POST` is more appropriate here. A `PUT` would normally imply that a new resource is being created at this path. Instead this endpoint performs an action and subsequent calls to it can produce different resources.

Another alternative would be to add `PUT("/configurations/:name", ...)` with a way to specify a source configuration (url or body param), but this would suggest that we create other configs this way (e.g. raw and parameterized) and I think we want to just keep using `/apply` for that.

model/configuration.go
```diff
@@ -482,3 +482,21 @@ func (rc *ResourceConfiguration) indexFields(resourceName string, resourceTypeNa
...
+// Duplicate copies the value of the current configuration and returns
+// a duplicate with the new name.  It should be identical except for the
+// Metadata.Name, Metadata.ID, and Spec.Selector fields.
+func (c *Configuration) Duplicate(name string) *Configuration {
+	new := &Configuration{}
+	*new = *c
+
+	// Change the metadata values
+	new.Metadata.Name = name
+	new.Metadata.ID = name
```
andykellr, line 495:
```suggestion
	new.Metadata.ID = uuid.NewString()
```

model/configuration.go
```diff
@@ -482,3 +482,21 @@ func (rc *ResourceConfiguration) indexFields(resourceName string, resourceTypeNa
 	index(resourceName, rc.Name)
 	index(resourceTypeName, rc.Type)
 }
+
+// Duplicate copies the value of the current configuration and returns
+// a duplicate with the new name.  It should be identical except for the
+// Metadata.Name, Metadata.ID, and Spec.Selector fields.
+func (c *Configuration) Duplicate(name string) *Configuration {
+	new := &Configuration{}
+	*new = *c
```
andykellr, line 491: Even though it is legal go, shadowing the builtin `new` function is confusing here. Call it `copy` instead.

Also, this is sufficient:

```go
copy := *c
```

And then change the return to

```go
return &copy
```

internal/rest/api.go
dsvanlani: Ok - sounds like this should just be a POST.

andykellr (changes requested):
Looks good other than those few comments.

andykellr (approved):

PR #162, opened 2022-07-27
Set 60 second timeout for test cases

jsirianni:
### Proposed Change

It is hard to track down flaky tests when not using a timeout. 60 seconds should be more than enough for the current tests, and short enough to catch things quicker.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


dsvanlani (approved):
LGTM

PR #168, opened 2022-07-28
feat(destination-type): Splunk Observability Cloud (SignalFx)

jsirianni:
### Proposed Change

***NOTE***: observiq-otel-collector does not support the exporters required by this destination type. This PR should not be merged before support is added to the collector. I will have a PR adding the exporters tomorrow.

Splunk Observability Cloud (SignalFX) requires the following exporters
- SAPM: traces
- Splunk HEC: logs
- SignalFX: metrics

The token for all three telemetry types appears to be the same, however, the platform can enable / disable features (logs will not work by default). Because of this, the destination type has toggles for each telemetry type.

The resource detection processor is **required** per the signalfx exporter docs.

I was able to get trace and metrics working without issue, however, logs gave 403 errors. This is because I do not have logging enabled, and the Splunk UI indicates that I need a Splunk Enterprise account when I try to setup logging. I am confident that the configuration is correct as it is derived from the configuration given to me during a splunk collector install.

```yaml
# splunk log exporter config generated by Splunk's platform
...
  splunk_hec:
    token: "<my token>"
    endpoint: "https://ingest.us1.signalfx.com/v1/log"
    source: "otel"
    sourcetype: "otel"
...
```

Screenshot from 2022-07-27 21-24-14

Metrics

Screenshot from 2022-07-27 20-27-38

Traces

Screenshot from 2022-07-27 20-09-55

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


jsirianni:
The collector will need to merge this https://github.com/observIQ/observiq-otel-collector/pull/593. We will also want to wait for a release.

andykellr (approved):

resources/destination-types/signalfx.yaml
```diff
@@ -0,0 +1,89 @@
+apiVersion: bindplane.observiq.com/v1beta
```
dsvanlani, line 1:
```suggestion
apiVersion: bindplane.observiq.com/v1
```

dsvanlani (changes requested):

PR #172, opened 2022-07-28
feat(source-type): Kafka Cluster metrics

jsirianni:
### Proposed Change

Using the `kafkametrics` receiver, we can get cluster level metrics such as "number of brokers" and metrics related to topics / partitions.

I will have a second PR adding a kafka source type which will use JMX for metrics and the Kafka plugin for logs.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


resources/source-types/kafka_cluster.yaml
```diff
@@ -0,0 +1,294 @@
...
+    - name: cluster_name
+      label: Cluster Name
+      description: Friendly name used for the resource kafka.cluster.name.
+      type: string
+      default: ""
+      required: true
+
+    - name: protocol_version
+      label: Protocol Version
+      description: Protocol version.
```
andykellr, line 24: We need a better description here. See guidelines in https://github.com/observIQ/stanza-plugins

resources/source-types/kafka_cluster.yaml
```diff
@@ -0,0 +1,294 @@
...
+      advancedConfig: true
+
+    - name: auth_type
+      label: Auth Type
+      type: enum
+      validValues:
+        - basic
+        - sasl
+        # TODO: Should be tested before implemented into the destination type
+        #- tls
```
andykellr, line 65: This is a source-type. Copy/paste? Something else we're waiting for?

jsirianni: Yes, the auth stuff is shared between kafka metrics, kafka receiver, kafka exporter. I will update.

The TLS config options looked suspicious in the receiver documentation. I would like to test them before assuming they work, but I do not have an environment available right now.

resources/source-types/kafka_cluster.yaml
jsirianni: I am tempted to call it a "minimum" supported version but that is not really what it is (unlike TLS, which will negotiate the highest supported version on both ends). 

I added a couple more options, even though 2.0.0 tends to work on newer clusters. It is possible a user might have their cluster configured with a min version of 2.2.1 (latest).

andykellr (approved):
We could wait for the new control or merge as is and update later.

PR #173, opened 2022-07-28
feat(source-type: Kafka Node JMX metrcis and logs

jsirianni:
### Proposed Change

Kafka node source type gathers JMX metrics and logs from a single Kafka node.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


andykellr (approved):
Since this doesn't include the icon, let's make sure #172 is merged first.

PR #174, opened 2022-07-29
feat(metric filtering): Rabbitmq metric toggles

jsirianni:
### Proposed Change

Added toggles for disabling Rabbitmq metrics. The list of metrics was referenced here https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/rabbitmqreceiver/metadata.yaml.

Screenshot from 2022-07-28 20-24-25

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/resources/source-types/rabbitmq.yaml
+++ b/resources/source-types/rabbitmq.yaml
@@ -63,6 +63,35 @@ spec:
           value: true
       advancedConfig: true
 
+    - name: metric_filtering
+      label: Metric Filtering
+      type: bool
+      default: false
+      relevantIf:
+        - name: enable_metrics
+          operator: equals
+          value: true
+      advancedConfig: true
+
+    - name: disable_metrics
+      label: Disable Metrics
+      type: enums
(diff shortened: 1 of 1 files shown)


andykellr:
The heading `Disable metrics` makes the toggles confusing. By turning on the toggle, are you disabling them or enabling them?

andykellr:
@dsvanlani I think we need to tweak the CSS on the new `enums` field to indent.

jsirianni:
> The heading `Disable metrics` makes the toggles confusing. By turning on the toggle, are you disabling them or enabling them?

I agree. I would rather have a metrics toggle for "enabling" them, with all metrics enabled by default.

I think a custom function that takes a list of metrics to be enabled (and knows of the full list) would be nice. It could return the enable metrics as "enabled" and the remaining metrics as "disabled".

jsirianni:
Superseded by https://github.com/observIQ/bindplane-op/pull/216

PR #176, opened 2022-07-29
feat(processor): Custom processor

jsirianni:
### Proposed Change

The custom processor allows the user to define a processor using the `yaml` parameter type.

TODO: I think we should have toggles for each pipeline, instead of assuming the processor should be used for every pipeline.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


dsvanlani:
Closing in favor of #204

PR #181, opened 2022-08-01
feat: Parameter option "creatable" for enum parameter.

dsvanlani:
### Proposed Change
The `Parameter` model now has an `options` field which is meant to be used to customize parameters in the UI.  The first supported options is `creatable` for the `enum` type.

```go
// ParameterOptions specify further customization for inputs in the UI
type ParameterOptions struct {
	// Creatable will modify the "enum" parameter from a select to
	// a creatable select where a user can specify a custom value
	Creatable bool `json:"creatable" yaml:"creatable"`
}
```

We have an example of this in in `model/testfiles/sourcetype-experimental-parameters.yaml`

```yaml
 parameters:
    - name: creatable_enum_param
      type: enum
      label: Creatable Enum
      required: true
      description: '"BYOV" - bring your own value.'
      validValues:
        - one
        - two
        - three
        - four
      options:
        creatable: true
```

You can apply it with `go run cmd/bindplanectl/main.go apply -f model/testfiles/sourcetype-experimental-parameters.yaml`.

And in the UI we are using the `MUI Autocomplete` component to show a text input with suggested options.

image

image

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/model/parameter.go
+++ b/model/parameter.go
@@ -49,7 +49,6 @@ type ParameterDefinition struct {
 	Type string `json:"type" yaml:"type"`
 
 	// only useable if Type == "enum"
-	Creatable   bool     `json:"creatable" yaml:"creatable"`
 	ValidValues []string `json:"validValues,omitempty" yaml:"validValues,omitempty" mapstructure:"validValues"`
 
 	// Must be valid according to Type & ValidValues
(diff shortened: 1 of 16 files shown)


model/parameter.go
```diff
@@ -56,6 +56,15 @@ type ParameterDefinition struct {
 	RelevantIf     []RelevantIfCondition `json:"relevantIf,omitempty" yaml:"relevantIf,omitempty" mapstructure:"relevantIf"`
 	Hidden         bool                  `json:"hidden" yaml:"hidden"`
 	AdvancedConfig bool                  `json:"advancedConfig" yaml:"advancedConfig" mapstructure:"advancedConfig"`
+
+	Options ParameterOptions `json:"options" yaml:"options"`
+}
+
+// ParameterOptions specify further customization for inputs in the UI
```
jsirianni, line 63: nit: Technically this will work anywhere, right? Not just the ui.

```suggestion
// ParameterOptions specify further customization for input parameters
```

dsvanlani: True.  This would change the validation for an `enum`

jsirianni (approved):

PR #200, opened 2022-08-01
feat(source-type): SAP Hana logs source type

jsirianni:
### Proposed Change

Added SAP Hana source type with logs only. Metrics is commented out, as we need to wait for a collector release that includes the Hana receiver.

Screenshot from 2022-08-03 14-15-28

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


dsvanlani (changes requested):
Nit: I think if metrics support is not in we should not have the relevant_if for logs.

image

image


Config rendering looks good.

```yaml
receivers:
    plugin/hana__source0:
        parameters:
            file_path:
                - /usr/sap/*/HDB*/*/trace/*.trc
            start_at: end
            timezone: UTC
        path: $OIQ_OTEL_COLLECTOR_HOME/plugins/sap_hana_logs.yaml
processors:
    batch/otlp_grpc__otlp: null
    resourcedetection/hana__source0:
        detectors:
            - system
        system:
            hostname_sources:
                - os
exporters:
    otlp/otlp_grpc__otlp:
        endpoint: 0.0.0.0:4317
        tls:
            insecure: true
service:
    pipelines:
        logs/hana__source0__otlp:
            receivers:
                - plugin/hana__source0
            processors:
                - resourcedetection/hana__source0
                - batch/otlp_grpc__otlp
            exporters:
                - otlp/otlp_grpc__otlp
```

jsirianni:
@dsvanlani totally agree, done.

resources/source-types/sap_hana.yaml
```diff
@@ -0,0 +1,269 @@
+apiVersion: bindplane.observiq.com/v1beta
+kind: SourceType
+metadata:
+  name: hana
+  displayName: SAP Hana
+  icon: /icons/sources/sap.svg
+  #description: SAP Hana metrics and logs
+  description: SAP Hana logs
```
ryancgoins, line 8:
```suggestion
  description: Collect logs from SAP Hana 
```

I'm working on a updating our Stanza style guide for us to use with BindPlane OP so we have a standard here. For now, let's use the above format for descriptions

ryancgoins (approved):

PR #202, opened 2022-08-01
feat(CLI): Copy configuration command.

dsvanlani:
### Proposed Change
This PR adds a command `bindplanectl copy configuration <name of config> <name of copy>` to the bindplanectl CLI.

Also valid:
- `bindplanectl cp configuration <name> <name>`
- `bindplanectl copy config <name> <name`

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


client/bindplane.go
```diff
@@ -182,11 +184,11 @@ type bindplaneClient struct {
 var _ BindPlane = (*bindplaneClient)(nil)
 
 // NewBindPlane takes a client configuration, logger and returns a new BindPlane.
-func NewBindPlane(config *common.Client, logger *zap.Logger) (BindPlane, error) {
+func NewBindPlane(config *common.Client, logger *zap.Logger, url string) (BindPlane, error) {
```
jsirianni, line 187: This is a breaking change to the client public api. I might be missing something, but I think the motivation here is for test cases. Could the test modify the URL after calling NewBindPlane? We could add a set method perhaps?

dsvanlani: Thats correct.  I think that would make sense.  I'll refactor now and re enable the integration tests.

PR #177, opened 2022-07-29
feat(processor): Add / Delete Attributes / Resources

jsirianni:
### Proposed Change

Processors for adding and deleting attributes / resources.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

PR #205, opened 2022-08-03
feat(processor): Add / Delete Attributes / Resources

dsvanlani:
### Proposed Change
Cleaned up version of #177 
- Adds 4 new processors: add_attribute, add_resource, delete_attribute, delete_resource

With this processor section:

image

I was able to produce this config:

```yaml
receivers:
    plugin/file__source0:
        parameters:
            encoding: utf-8
            file_path:
                - /tmp/file.log
            log_type: file
            multiline_line_start_pattern: ""
            parse_format: none
            start_at: end
        path: $OIQ_OTEL_COLLECTOR_HOME/plugins/file_logs.yaml
processors:
    attributes/add_attribute__file__source0__processor1:
        actions:
            - action: upsert
              key: computer
              value: daves-macbook-pro
            - action: upsert
              key: something
              value: blah
    attributes/delete_attribute__file__source0__processor2:
        actions:
            - action: delete
              key: one
            - action: delete
              key: two
            - action: delete
              key: three
    resource/add_resource__file__source0__processor3:
        attributes:
            - action: upsert
              key: blah
              value: baz
            - action: upsert
              key: foo
              value: bar
    resource/delete_resource__file__source0__processor0:
        attributes:
            - action: delete
              key: blah
            - action: delete
              key: foo
            - action: delete
              key: bar
    resourcedetection/file__source0:
        detectors:
            - system
        system:
            hostname_sources:
                - os
exporters:
    logging/logging__info:
        logLevel: info
service:
    pipelines:
        logs/file__source0__info:
            receivers:
                - plugin/file__source0
            processors:
                - resourcedetection/file__source0
                - resource/delete_resource__file__source0__processor0
                - attributes/add_attribute__file__source0__processor1
                - attributes/delete_attribute__file__source0__processor2
                - resource/add_resource__file__source0__processor3
            exporters:
                - logging/logging__info
```

With these processor configurations:

**Add Attribute**
image

**Add Resource**
image

**Delete Attribute**
image

**Delete Resource 
image

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


resources/processor-types/add_attribute.yaml
```diff
@@ -0,0 +1,71 @@
+apiVersion: bindplane.observiq.com/v1beta
+kind: ProcessorType
+metadata:
+  name: add_attribute
+  displayName: Add Attribute
+spec:
+  version: 0.0.1
+  parameters:
+    - name: attributes
+      label: Map Type Param
```
jsirianni, line 10:
```suggestion
      label: Attributes
```

resources/processor-types/add_resource.yaml
```diff
@@ -0,0 +1,71 @@
+apiVersion: bindplane.observiq.com/v1beta
+kind: ProcessorType
+metadata:
+  name: add_resource
+  displayName: Add Resource
+spec:
+  version: 0.0.1
+  parameters:
+    - name: resources
+      label: Map Type Param
```
jsirianni, line 10:
```suggestion
      label: Resources
```

jsirianni (changes requested):

jsirianni (approved):

resources/processor-types/add_attribute.yaml
```diff
@@ -0,0 +1,71 @@
...
+  displayName: Add Attribute
+spec:
+  version: 0.0.1
+  parameters:
+    - name: attributes
+      label: Attributes
+      type: map
+      required: true
+
+    - name: action
```
ryancgoins, line 14: I think we should move the action above the parameter. Seems more natural to select that first and then enter the key/value pairs

resources/processor-types/add_attribute.yaml
```diff
@@ -0,0 +1,71 @@
...
+      label: Action
+      type: enum
+      description: 'insert: Add attribute if it does not exist. update: Update existing value. upsert: Insert or update.'
+      default: upsert
+      validValues:
+        - insert
+        - update
+        - upsert
+
+    - name: enable_metrics
```
ryancgoins, line 24: Can we add a description to this section that says "Select which types of telemetry you'd like to enable this processor for."

ryancgoins (commented):
Added some minor thoughts on copy / layout. Otherwise looks good

resources/processor-types/add_attribute.yaml
dsvanlani: This could be a `multi-enum` parameter type.

resources/processor-types/add_attribute.yaml
jsirianni: I pushed a change to move action to the top.

Screenshot from 2022-08-03 21-57-02

resources/processor-types/add_attribute.yaml
jsirianni: Bool values do not display their descriptions currently. 

We could do a multi select enum, which would have a single description with all three pipelines as options in the select. It would make the template somewhat more complicated. We would need to loop over the values, checking to see if the value is <pipeline type> and then use it.

dsvanlani:
@ryancgoins This is the new layout for the Telemetry section for these processors:

image

ryancgoins:
LGTM

PR #208, opened 2022-08-03
feat(UI): Handle enter keys inside the `map` Parameter Input

dsvanlani:
### Proposed Change
After using the `map` type parameter input on recent PRs I realized that we do not handle the `Enter` key very well (at all).

It's pretty natural to hit enter to go to the next value - or at the end of your input and since we were not overriding the default form submission on enter this could mean a user didn't actually save the last value for the field.

Basically:  Hitting enter in a input will go to the next input or add a row.  It feels much more natural.

### To test

Make sure you have the Experimental sourcetype

```sh
bindplanectl apply -f model/testfiles/sourcetype-experimental-parameters.yaml
```

https://user-images.githubusercontent.com/64915094/182682467-a49fd985-19e4-4b0e-af60-68600ec8bf83.mov

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


jsirianni:
Working well for me outside of the test failures.

jsirianni (approved):
:+1:

PR #210, opened 2022-08-04
fix: Clear `StringsInput` editing value on `blur`.

dsvanlani:
### Proposed Change
Fixed a bug where the strings input would not correctly set the input value on "blur".  That is, when a user clicked away from the strings input *without* hitting enter the value was saved as a pill while not clearing the editing value making it appear twice in the input.

screenshot_from_2022-08-03_17-46-16

Now you can click away and it saves the value correctly:

https://user-images.githubusercontent.com/64915094/182735556-54e05023-7db4-42fb-ac1f-6977c3241167.mov

##### Checklist
- [x] Changes are tested
- [x] CI has passed


andykellr (approved):
I noticed this the other day when I hit TAB after entering a value. Good fix!

jsirianni (approved):
Working great for me, can no longer re-produce this.

PR #214, opened 2022-08-04
feat: `timezone` Parameter Type

dsvanlani:
### Proposed Change
- Adds a parameter type `timezone` to simplify Resource Type authoring + Allow for a custom control.
- Uses the MUI Autocomplete so a user can type in the box to find a desired timezone.
- Updates several source-types to use this new parameter type.

https://user-images.githubusercontent.com/64915094/182913966-e74315dd-ef37-4920-9eda-cd86fac801dc.mov

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


model/parameter.go
```diff
@@ -347,6 +350,23 @@ func (p ParameterDefinition) validateEnumsValue(fieldType parameterFieldType, va
 	return err.ErrorOrNil()
 }
 
+func (p ParameterDefinition) validateTimezoneType(fieldType parameterFieldType, value any) error {
+	tzErr := errors.NewError(fmt.Sprintf("invalid value for timezone for parameter %s", p.Name),
```
jsirianni, line 354: nit: Would it make sense to replace NewError and Sprintf with fmt.Errorf?

model/parameter.go
```diff
@@ -382,3 +402,22 @@ func (p ParameterDefinition) validateMapValue(fieldType parameterFieldType, valu
...
+// isValid returns true if str is contained in the list of timezone values
+func (t *tz) isValid(str string) bool {
+	for _, v := range t.values {
+		if v == str {
+			return true
+		}
+	}
+
+	return false
+}
```
jsirianni, line 423: It might be more efficient to use a map here, allowing you to check for an index instead of iterating. See the collector's implementation https://github.com/observIQ/observiq-otel-collector/blob/main/receiver/pluginreceiver/timezone.go#L17

model/parameter.go
dsvanlani: NewError is nice because it allows for a second argument as a suggestion to resolve the error and its what we've been using in this package so far.

model/parameter.go
dsvanlani: Sure would, I wonder if it would be worth asking them to export that Timezone map in the future.

jsirianni (approved):

jsirianni (approved):

PR #216, opened 2022-08-04
feat: RabbitMQ metric toggles

dsvanlani:
### Proposed Change
This PR adds a metric toggles section to the RabbitMQ source.

To do this we added another value in `ParameterOptions` - `TrackUnchecked`

```go
type ParameterOptions struct {
	// Creatable will modify the "enum" parameter from a select to
	// a creatable select where a user can specify a custom value
	Creatable bool `json:"creatable" yaml:"creatable"`

	// TrackUnchecked will modify the "enums" parameter to store the
	// unchecked values as the value.
	TrackUnchecked bool `json:"trackUnchecked" yaml:"trackUnchecked"`
}
```

This applies to the `enums` param and implies that we store the value of unchecked items rather than checked.

I.e the value value of this parameter is stored as 
```
value: ["rabitmq.consumer.count"]
```

image

Which is then used to produce this receivers section of an OTel config:

```yaml
receivers:
    plugin/rabbitmq__source0__rabbitmq:
        parameters:
            daemon_log_paths:
                - /var/log/rabbitmq/rabbit@*.log
            start_at: end
        path: $OIQ_OTEL_COLLECTOR_HOME/plugins/rabbitmq_logs.yaml
    rabbitmq/rabbitmq__source0:
        collection_interval: 60s
        endpoint: http://localhost:15672
        metrics:
            rabbitmq.consumer.count:
                enabled: false
        password: admin
        tls:
            insecure: true
        username: admin

```

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


jsirianni (commented):
Working well for me. Before using the toggle, I was getting five metrics
```
{"level":"info","ts":"2022-08-05T11:50:01.688-0400","msg":"MetricsExporter","kind":"exporter","data_type":"metrics","name":"logging/logging__info","#metrics":5}
```
I disabled two metrics and now I see three
```
{"level":"info","ts":"2022-08-05T11:53:06.286-0400","msg":"MetricsExporter","kind":"exporter","data_type":"metrics","name":"logging/logging__info","#metrics":3}
```
I have a dashboard showing this in real time
Screenshot from 2022-08-05 11-55-37

Screenshot from 2022-08-05 11-55-54

jsirianni (approved):

andykellr (approved):

PR #227, opened 2022-08-06
fix: Show Processor Telemetry Types 

dsvanlani:
### Proposed Change
- Query for telemetry types and pass them through the Resource Button in the Processor Select View.

image

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


andykellr (approved):
I think add/delete attribute is logs only, but we should confirm with @jsirianni

jsirianni:
> I think add/delete attribute is logs only, but we should confirm with @jsirianni

It is. I have a fix coming in a PR. I think it is fine for now.

jsirianni:
https://github.com/observIQ/bindplane-op/pull/228

dsvanlani:
Working correctly after merging Joes fix from main.
image

PR #251, opened 2022-08-09
Sync agent-versions at regular intervals

andykellr:
### Proposed Change
Adds a goroutine to sync agent-version resources with GitHu at regular intervals.

The following command can be used to change the interval:

```
bindplanectl profile set local --sync-agent-versions-interval 6h
```

The default and minimum interval is 1h.

This PR also adds a listener for AgentVersion resources changes and automatically resets the LatestVersion cache when there is an AgentVersion change.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


internal/agent/versions.go
```diff
@@ -62,13 +71,22 @@ var _ Versions = (*versions)(nil)
...
+func NewVersions(ctx context.Context, client Client, store store.Store, settings VersionsSettings) Versions {
+	v := &versions{
 		client:        client,
 		store:         store,
 		latestVersion: util.NewRemember[model.AgentVersion](latestVersionCacheDuration),
 		logger:        settings.Logger,
 	}
+	if settings.SyncAgentVersionsInterval > 0 && !settings.Offline {
+		interval := settings.SyncAgentVersionsInterval
+		if interval < time.Hour {
```
dsvanlani, line 83: This Got me - I set the interval to 10s and couldn't figure out what was going wrong.  We may want to mention this somewhere.

dsvanlani (changes requested):
One comment about a debug log.

Also we should mention in the command help (or in the `profile set` validation) that the minimum for the Sync interval is 1h.

internal/cli/commands/serve/start.go
```diff
@@ -248,8 +248,11 @@ func (s *Server) createVersions(config *common.Server, st store.Store) agent.Ver
 	if !config.Offline {
 		client = agent.NewClient()
 	}
-	return agent.NewVersions(client, st, agent.VersionsSettings{
-		Logger: s.logger.Named("versions"),
+	fmt.Println(config.SyncAgentVersionsInterval)
```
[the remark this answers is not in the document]
andykellr: Oh, we can just remove that.

internal/agent/versions.go
andykellr: It's here. Not sure where else to put it.

```
go run ./cmd/bindplanectl profile set local --help
set a parameter on a saved profile

Usage:
  bindplanectl profile set <name> [flags]

Flags:
      --disable-downloads-cache                 true if agent distributions should be cached
      --downloads-folder-path string            full path to the downloads folder where agents are cached, defaults to $HOME/.bindplane/downloads
  -h, --help                                    help for set
      --offline                                 BindPlane server should be run in offline mode
      --remote-url string                       websocket url that agents use to connect to the server
      --secret-key string                       secret key used by agents when connecting to the server
      --sessions-secret string                  secret key used to sign cookies for session authentication, must be a UUID
      --storage-file-path string                full path to the desired storage file, defaults to the $HOME/.bindplane/storage
      --store-type string                       type of store to use for storing agent status and configuration resources
      --sync-agent-versions-interval duration   time interval to sync agent-version resources from GitHub releases, 0 to disable or minimum 1h (default 1h0m0s)

Global Flags:
  -c, --config string          full path to configuration file
      --env string             BindPlane environment. One of test|development|production (default "production")
      --host string            domain on which the BindPlane server will run (default "localhost")
      --log-file-path string   full path of the BindPlane log file, defaults to $HOME/.bindplane/bindplane.log
      --log-output string      output of the log. One of: file|stdout
  -o, --output string          output format. One of: json|table|yaml|raw (default "table")
      --password string        password to use with Basic auth (default "admin")
      --port string            port on which the rest server is listening (default "3001")
      --profile string         configuration profile name to use
      --server-url string      http url that clients use to connect to the server
      --tls-ca strings         TLS certificate authority file(s) for mutual TLS authentication
      --tls-cert string        TLS certificate file
      --tls-key string         TLS private key file
      --tls-skip-verify        Whether to verify the server's certificate chain and host name when making client requests
      --username string        username to use with Basic auth (default "admin")
```

dsvanlani: Sounds good 👍

dsvanlani (approved):

PR #290, opened 2022-08-17
feat: Agents Count field on Configuration Table

dsvanlani:
### Proposed Change
This adds an Agents column to the configuration table where we show the number of agents using that config.

- Adds a field to the Configuration GraphQL model `agentCount`
- Alters the `ConfigurationChanges` query to include configurations when agents are labeled, inserted, or removed.

Resolves observiq/iris#1140

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


andykellr (approved):

internal/graphql/schema.resolvers.go
```diff
@@ -90,6 +90,13 @@ func (r *configurationResolver) Kind(ctx context.Context, obj *model.Configurati
 	return string(obj.GetKind()), nil
 }
 
+// AgentCount is the resolver for the agentCount field.
+func (r *configurationResolver) AgentCount(ctx context.Context, obj *model.Configuration) (*int, error) {
```
jmwilliams89, line 94: Why do we return a pointer to an integer?

internal/graphql/schema.resolvers.go
```diff
@@ -90,6 +90,13 @@ func (r *configurationResolver) Kind(ctx context.Context, obj *model.Configurati
...
+	ids, err := r.bindplane.Store().AgentsIDsMatchingConfiguration(obj)
```
jmwilliams89, line 95: Related to above. If we're comfortable returning a pointer with this function, we should probably return nil for the count if there is an error.

internal/graphql/schema.resolvers.go
andykellr: This is an optional value (no !), so it returns a pointer. It's not really optional but it helps with the types in the client. We will have another PR which changes this and fixes up the types in the client (lots of little, mostly meaningless changes). At this point this will no longer be a pointer.

jmwilliams89 (approved):

PR #293, opened 2022-08-18
feat: Filter processor types by supported telemetry.

dsvanlani:
### Proposed Change
In the Processor Select View make sure we show only processors that can be used with the given source.  That is, if a source only supports metrics - do not show processors that are log only like `Delete Attribute` etc.

Consider the Host source:
image

Misc:
- Fixed some tests mocks for the Configuration table tests
- Fixed a eslint warning in the SourcesSection component

Resolves: https://github.com/observIQ/iris/issues/1132

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


jsirianni (approved):
Working well for me.

Screen Shot 2022-08-18 at 2 19 33 PM

andykellr (approved):

Issue #273, opened 2022-08-12
Verify that Metric filtering metrics are all valid (e.g no sub metrics)


andykellr:
Some metrics were deprecated and should not be in the metric list.

For example:

```yaml
  # produced when receiver.hostmetricsreceiver.emitMetricsWithDirectionAttribute feature gate is enabled
  system.network.packets:
    enabled: true
    description: The number of packets transferred. (Deprecated)
    unit: "{packets}"
    sum:
      value_type: int
      aggregation: cumulative
      monotonic: true
    attributes: [device, direction]

  # produced when receiver.hostmetricsreceiver.emitMetricsWithoutDirectionAttribute feature gate is enabled
  system.network.packets.transmit:
    enabled: true
    description: The number of packets transmitted.
    unit: "{packets}"
    sum:
      value_type: int
      aggregation: cumulative
      monotonic: true
    attributes: [device]

  # produced when receiver.hostmetricsreceiver.emitMetricsWithoutDirectionAttribute feature gate is enabled
  system.network.packets.receive:
    enabled: true
    description: The number of packets received.
    unit: "{packets}"
    sum:
      value_type: int
      aggregation: cumulative
      monotonic: true
    attributes: [device]
```

We should remove the deprecated metrics from the filters and check all sources for similar issues.

PR #295, opened 2022-08-18
fix: Remove feature gated metrics from existing metric toggles.

dsvanlani:
### Proposed Change
This PR removes metrics that rely on the feature gate `emitMetricsWithDirectionAttribute` to be enabled.

A discussion on this can be found in the VCenter receiver readme in the otel collector contribe [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/vcenterreceiver).

**This is a breaking change for anyone using this feature**

Right now we validate that a source configuration values for `parameters` are one of `validValues`.  We are removing valid values from this list so if someone had say `system.network.packets.receive` disabled with the metric toggles then their source would no longer be valid.

To fix this the customer would need to delete the source and reconfigure it.

Resolves: https://github.com/observIQ/bindplane-op/issues/273

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


resources/source-types/elasticsearch.yaml
```diff
@@ -193,8 +193,6 @@ spec:
         - elasticsearch.node.cache.memory.usage
         - elasticsearch.node.cluster.connections
         - elasticsearch.node.cluster.io
-        - elasticsearch.node.disk.io.read
-        - elasticsearch.node.disk.io.write
```
andykellr, line 197: The cluster.io metrics are feature gated, but oddly it doesn't seem that the disk.io metrics are affected.

andykellr (changes requested):
elasticsearch disk metrics should probably remain.

andykellr (approved):

jsirianni (approved):

PR #341, opened 2022-09-06
feat(ui): Resource Configuration Form Design + Metrics Parameter Type

dsvanlani:
Resolves: https://github.com/observIQ/iris/issues/1135
Resolves: https://github.com/observIQ/iris/issues/1148,

### Proposed Change
Implements a `metric` parameter type for metric toggles in Sources.  Check out the figma design [here](https://www.figma.com/file/aO8gL7nXYIwdYYfSpybjdW/Metrics-pop-up?node-id=0%3A1).

#### Resource Form Changes

- Parameters are shown over two columns.
- Increased size of Title and description.
- "Action" buttons and Title are always visible in the Dialog - the content inside scrolls.

https://user-images.githubusercontent.com/64915094/188720640-693d9759-2020-45e0-94de-f43d4ee61646.mov

- Hitting Save is always possible for new resources, and on click will validate all fields and show errors.

https://user-images.githubusercontent.com/64915094/188719710-4d827212-4425-4c37-b9d8-88d3d90f846d.mov

#### Metrics Parameter Type

- New parameter type "metrics" which takes in metric categories as options.  See definitions in `parameter.go`.

#### New fields for Options on the paramter model
Grid columns - specifies the number of flex columns the parameter should span, a number between 1 and 12
Section Header - used to signal the UI to embolden the text for a Boolean parameter that acts as section head.
Metric Categories - used for the metrics input.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


resources/source-types/iis.yaml
```diff
@@ -33,6 +33,8 @@ spec:
       label: Metric Filtering
```
Mrod1598, line 33: All the other sources say just `Filtering` here.

resources/source-types/hbase.yaml
```diff
@@ -56,6 +56,8 @@ spec:
       description: Enable to scrape master server's JMX port.
```
Mrod1598, line 56: Would it be a meaningful change to add a disclaimer to the ones that don't have metric filtering? not sure if we expect users to click through and checkout multiple sources but if going from one with filtering and one without filtering it might be confusing why some don't support filtering and some do.

resources/source-types/aerospike.yaml
```diff
@@ -71,6 +71,8 @@ spec:
         - name: enable_metrics
```
Mrod1598, line 71: do we do any validation on hostname/port or do we want to?
image

Mrod1598 (approved):

Mrod1598 (changes requested):
Looks good over all with a few comments ^(misclicked)

andykellr:
I noticed a strange issue testing the macOS source. There is a `Start At` field that is probably supposed to be part of the System Logs section. If I toggle `Enable Logs` on and off, another field is added.

https://user-images.githubusercontent.com/2660251/189019086-6b98e517-41f7-4b70-9d9b-d0f82d2ec95d.mov

model/parameter.go
```diff
@@ -79,6 +80,33 @@ type ParameterOptions struct {
...
 	TrackUnchecked bool `json:"trackUnchecked" yaml:"trackUnchecked"`
+
+	// GridColumns will specify the number of flex-grid columns the
+	// control will take up, must be an integer between 1 and 12 or
+	// unspecified.
+	GridColumns *int `json:"gridColumns,omitempty" yaml:"gridColumns,omitempty"`
+
+	// SectionHeader is used to indicate that the bool parameter input is
+	// a switch for further configuration for UI styling.
+	SectionHeader *bool `json:"sectionHeader,omitempty" yaml:"gridColumns,omitempty"`
```
andykellr, line 91:
```suggestion
	SectionHeader *bool `json:"sectionHeader,omitempty" yaml:"sectionHeader,omitempty"`
```

model/parameter.go
```diff
@@ -228,6 +256,82 @@ func (p ParameterDefinition) validateOptions() error {
...
+			)
+
+			return errs
+		}
+
+		for _, category := range p.Options.MetricCategories {
+			err := category.validateMetricCategory()
+			if err != nil {
+				multierror.Append(errs, err)
+			}
```
andykellr, line 291: Even if you add to errs here, it always returns nil.

I would recommend refactoring validateOptions and validateMetricCategories to take `errs validation.Errors` (which is built on multierror) and add them directly there.

We could consider refactoring the other methods of validateDefinition similarly but I left them alone because they all only report a single error (or nil).

model/parameter_test.go
```diff
@@ -531,3 +538,69 @@ eleven:
...
+				Metrics: []MetricOption{{}, {KPI: boolPtr}, {Description: descPtr}},
+			},
+			true,
+		},
+	}
+
+	for _, test := range testCases {
+		err := test.metricCategory.validateMetricCategory()
+		if test.expectErr {
+			require.Error(t, err)
```
andykellr, line 601: I think we should validate the text of the error.

Also, it would be good to include in `validation_test.go` a few examples of these test cases with actual yaml files (in the testfiles/validate folder)

resources/source-types/aerospike.yaml
andykellr: The following issues mention this. I think considering that this behavior hasn't changed with this PR, it's ok to leave as is. We should consider allocating these issues if we think this is important.

observiq/iris#915 
observiq/iris#1168
observiq/iris#1169

ui/src/components/ResourceConfigForm/ParameterInput/MetricsParamInput.tsx
```diff
@@ -0,0 +1,168 @@
...
+                    onClick={(e) => handleDeselectAll(e, allValues)}
+                  >
+                    Deselect All
+                  </button>
+                  |
+                  <button
+                    className={styles["metric-category-button"]}
+                    onClick={(e) => handleSelectAll(e, allValues)}
+                  >
+                    Select All
```
andykellr, line 137: In bindplane-og we used `Enable All | Enable KPIs Only | Disable All`

We have a little less room here, but I wonder if we should use `None | KPIs | All`

andykellr (changes requested):
I think this generally looks really great besides a few pieces of feedback.

The video I posted is a real issue. The same thing happens with cassandra.

I'd like to load every source page before approving this, but I think I'll wait until the few known issues are addressed.

resources/source-types/host.yaml
```diff
@@ -24,16 +24,20 @@ spec:
...
+      type: metrics
       options:
-        trackUnchecked: true
+        metricCategories:
+          - label: Load Metrics
+            metrics:
+              - name: system.cpu.load_average.15m
+              - name: system.cpu.load_average.1m
+              - name: system.cpu.load_average.5m
       validValues:
```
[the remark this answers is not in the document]
dsvanlani: Good catch.

resources/source-types/hbase.yaml
dsvanlani: I'm gonna defer this one to @ryancgoins

ui/src/components/ResourceConfigForm/ParameterInput/MetricsParamInput.tsx
dsvanlani: I think that would work.  RIght now we dont support a `KPIs` button, in the future we would do this by adding a `kpi: true` field to the parameters like this:

```yaml
options:
  metricCategories:
    - label: Postgres Metrics
       metrics:
         - name: kpi.metric
           kpi: true
```

I can change the copy to Enable All / Disable All though, which I think is more clear.

resources/source-types/hbase.yaml
ryancgoins: Great call out @Mrod1598. I don't think we need this in the UI, but definitely in the documentation. I've created a ticket for the docs to be updated

Mrod1598, line 56: seems the hbase metric toggle has the same issue that the other one did.
image. its the master metrics toggle and the region metrics toggle.

resources/source-types/mysql.yaml
```diff
@@ -96,11 +96,27 @@ spec:
       type: bool
```
Mrod1598, line 96: mysql doing the multi `start_at` too 
image

resources/source-types/host.yaml
```diff
@@ -24,16 +24,19 @@ spec:
       label: Load Metrics
```
Mrod1598, line 24: host has some weird formatting
image

resources/source-types/tomcat.yaml
```diff
@@ -66,10 +66,37 @@ spec:
       type: bool
```
Mrod1598, line 66: tomcat has the same issue as others with the logs toggle.
image

Also, these two fields still show up when toggle is off.
image

resources/source-types/mysql.yaml
Mrod1598, line 96: start_at is in two spots that don't seem to make sense. (seems seperate from the duplication bug.)
image

Mrod1598 (changes requested):
found a couple more duplication bugs.

resources/source-types/mysql.yaml
dsvanlani: @Mrod1598 I went ahead and added in verification that we are not duplicating parameter names - which fixes all these issues.  

```go
func (s *ResourceTypeSpec) validateNoDuplicateParameterNames(errs validation.Errors) {
	// visited is a map of parameter names to bool
	visited := make(map[string]bool, 0)

	for _, p := range s.Parameters {
		if visited[p.Name] {
			errs.Add(
				errors.New(
					fmt.Sprintf("found multiple parameters with name %s", p.Name),
				),
			)
		} else {
			visited[p.Name] = true
		}
	}
}
```

resources/source-types/host.yaml
dsvanlani: Yeah, i agree I'm not sure what to do about this as it was not included in the design.  I discussed with @ryancgoins and we removed the copy for each metric parameter so it looks a bit better.

image

ryancgoins: Is there a reason we don't format it using the same "Filtering" setting and then everything exposed under that in 2 columns?

dsvanlani: Well - for backwards compatibility reasons each of the metrics is a separate parameter.  I can investigate how we could use them as one large metrics parameter.

Mrod1598 (approved):
LGTM

PR #348, opened 2022-09-07
fix(cli): Include [name] parameter as optional in `bindplanectl get config`

dsvanlani:
Resolves: https://github.com/observIQ/iris/issues/1141

- Adds [name] in the usage for `get configuration`
- changes all non-agent resources usage to `get <resource> [name]` instead of `get <resource> [id]`

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


andykellr (approved):

internal/cli/commands/get/configurations.go
```diff
@@ -26,7 +26,7 @@ import (
 // ConfigurationsCommand returns the BindPlane get configurations cobra command
 func ConfigurationsCommand(bindplane *cli.BindPlane) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:     "configurations",
+		Use:     "configurations [id]",
```
antonblock, line 29:
```suggestion
		Use:     "configurations [name]",
```

dsvanlani: Thank you!

antonblock (approved):

Issue #349, opened 2022-09-09
Components Table name field can include leading space.


dsvanlani:
For shorter names in the Components table, it can sometimes appear that they have a leading space.

<img width="388" alt="Screen Shot 2022-09-09 at 3 37 37 PM" src="https://user-images.githubusercontent.com/64915094/189434275-2ec58b76-b974-491e-8233-f954709fa169.png">

PR #350, opened 2022-09-09
Fix: Components table name field can have leading space.

dsvanlani:
Resolves: #349 

### Proposed Change
- Use a custom button component instead of the MUI button component which contains padding, style it to look like an anchor.
- While we're at it use the primary brand color summer-sky blue for all anchors.
- Fixed bug where there could be horizontal scroll when editing a destination. Set `maxWidth` on the editDestination Modal.

image

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


andykellr:
I think we can ditch the underlines on the links.

Screen Shot 2022-09-11 at 3 45 29 PM

Also, it appears that the font-size is a little smaller than the Kind text.

andykellr (changes requested):
See comment

antonblock (approved):

PR #359, opened 2022-09-12
fix(ui): Update cookie max age on each request to avoid kicking active users out.

dsvanlani:
Resolves: https://github.com/observIQ/iris/issues/1110

### Proposed Change
Previously we initialized the cookie max age as 60 minutes, so a user would get logged out no matter what after 60 minutes of use.

Now we will update the `MaxAge` of the cookie inside the `CheckSession` middleware.  This means that every time a session cookie is used to authenticate the request we will extend the cookies life by 15 minutes.

Notably we do *not* do this inside the `/verify` call that is performed every 60 seconds, if we did this here then a user would never be logged out after being idle.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


antonblock (approved):
Tested by reducing `MaxAge` to a minute and navigating around for a couple minutes

andykellr:
Looks good but away from my computer and can't test. Agree it would be good to have.

andykellr (approved):

PR #361, opened 2022-09-13
feat(processor-type): Log severity filter processor

jsirianni:
### Proposed Change

Added a processor type that wraps the filter processor, for filtering on log severity. The user can specify a "minimum" severity. All severities matching or exceeding the configured option will be passed through the pipeline. All other logs are dropped. If a severity is unknown, it is included. This was an opinionated choice, we could make it configurable if we want to. My feeling is that if a log has no severity or something random, it should just be included.

The filter processor does support custom mappings, to support user defined severity levels. We could consider this in the future. 

### Testing

You can test with the nginx source type. Set the access log to a file containing the following:

```
10.33.104.40 - - [11/Jan/2021:11:25:01 -0500] "GET / HTTP/1.1" 200 612 "-" "curl/7.58.0"
10.33.104.40 - - [11/Jan/2021:11:25:01 -0500] "GET /help HTTP/1.1" 404 178 "-" "curl/7.58.0"
10.33.104.40 - - [11/Jan/2021:11:25:01 -0500] "GET /about-us HTTP/1.1" 404 178 "-" "curl/7.58.0"
```

You can change the status codes to test different severity levels, as severity is derived from the http code.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


antonblock (approved):
Successfully tested using snapshots and the method described

andykellr (approved):

PR #387, opened 2022-09-20
feat: configuration topology

dsvanlani:
### Proposed Change

* Add a topology graph to the configuration page showing connections from sources to processors to destinations
* Add a new Overview Page with an overview graph showing connections from configuration to destinations
* Add measurement collection processors to rendered pipelines to show throughput measurements on the topology graph
* Add measurement columns to the agents and configurations tables

Resolves observiq/iris#1152
Resolves observiq/iris#1210
Resolves observiq/iris#1211
Resolves observiq/iris#1212
Resolves observiq/iris#1213
Resolves observiq/iris#1216

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


internal/graphql/overview.go
```diff
@@ -0,0 +1,98 @@
...
+			continue
+		}
+
+		// don't include configurations with no destinations
+		if len(c.Spec.Destinations) == 0 {
+			continue
+		}
+
+		// don't include configurations that aren't deployed to agents
+		agentIDs, err := store.AgentsIDsMatchingConfiguration(context.Background(), c)
```
antonblock, line 51:
```suggestion
		agentIDs, err := store.AgentsIDsMatchingConfiguration(ctx, c)
```

internal/graphql/schema.graphqls
```diff
@@ -200,8 +225,8 @@ enum PipelineType {
 }
 
 enum ParameterType {
-  string
-  strings
+  String
+  Strings
```
andykellr, line 229: I think I did this accidentally.

internal/graphql/schema.graphqls
```diff
@@ -344,10 +370,57 @@ type Snapshot {
...
+  # ID of the graph node for this metric
+  nodeID: String!
+  # logs, metrics, or traces
+  pipelineType: String!
+  # numeric value of the metric
+  value: Float!
+  # unit for the metric, e.g. B/s
+  unit: String!
+  # hack just for Samuel
+  agentID: ID
```
andykellr, line 407: @shazlehu Is this used?

internal/store/boltstore.go
```diff
@@ -970,6 +1005,434 @@ func deleteResource[R model.Resource](ctx context.Context, s *boltstore, kind mo
...
+}
+
+func getRollupFromPeriod(opts stats.QueryOptions) time.Duration {
+	switch opts.Period {
+	case 1 * time.Minute:
+		return 10 * time.Second
+	case 5 * time.Minute:
+		return 1 * time.Minute
+	case 1 * time.Hour, 24 * time.Hour:
+		return 5 * time.Minute
```
andykellr, line 1243: @dehaansa Is that value for 24 * time.Hour correct? I thought we were using 1 * time.Hour there.

internal/graphql/schema.graphqls
dsvanlani: I think `strings` and `string` is correct, this is the enum for the parameter types.

internal/graphql/schema.resolvers.go
```diff
@@ -498,6 +529,84 @@ func (r *subscriptionResolver) ConfigurationChanges(ctx context.Context, selecto
 	return channel, nil
 }
 
+
+func metricSubscriber(ctx context.Context, sendMetrics func(),  updateTicker *time.Ticker) {
```
andykellr: @shazlehu I like this refactoring, but we need to move utility methods into `resolver.go` or they will get removed when we run `make generate`

internal/graphql/schema.resolvers.go
```diff
@@ -37,6 +37,10 @@ import (
 	"go.uber.org/zap"
 )
 
+const CONFIGURATION_METRICS_UPDATE_INTERVAL = 10
+const OVERVIEW_METRICS_UPDATE_INTERVAL = 10
+const AGENT_METRICS_UPDATE_INTERVAL = 10
```
andykellr:
```suggestion
const CONFIGURATION_METRICS_UPDATE_INTERVAL = 10 * time.Second
const OVERVIEW_METRICS_UPDATE_INTERVAL = 10 * time.Second
const AGENT_METRICS_UPDATE_INTERVAL = 10 * time.Second
```

Make these the full time.Duration and update the code below that uses them.

internal/graphql/schema.graphqls
shazlehu: Yes, it's not a hack -- I thought I excised that comment. We need the agentID in the metric data for the Agents table to decide if the data applies to a specific agent's row.

internal/store/boltstore.go
dehaansa: Will fix.

internal/graphql/resolver.go
```diff
@@ -174,3 +182,181 @@ func (r *Resolver) hasAgentConfigurationChanges(updates *store.Updates) bool {
...
+	return resourceName
+}
+
+func overviewMetrics(ctx context.Context, bindplane server.BindPlane, period string) (*model1.GraphMetrics, error) {
+	if period == "" {
+		period = "1m"
+	}
+	d, err := time.ParseDuration(period)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse period %s with duration %s", period, d)
```
antonblock, line 208: I think `d` will always be 0 if an error is returned

internal/graphql/resolver.go
```diff
@@ -174,3 +182,181 @@ func (r *Resolver) hasAgentConfigurationChanges(updates *store.Updates) bool {
...
+
+	var graphMetrics []*model1.GraphMetric
+
+	returnMetrics := &model1.GraphMetrics{
+		Metrics: graphMetrics,
+	}
+
+	for _, id := range ids {
+		var singleID []string
+		singleID = append(singleID, id)
```
antonblock, line 316:
```suggestion
		singleID := []string{id}
```

internal/graphql/schema.resolvers.go
```diff
@@ -498,6 +535,82 @@ func (r *subscriptionResolver) ConfigurationChanges(ctx context.Context, selecto
 	return channel, nil
 }
 
+func metricSubscriber(ctx context.Context, sendMetrics func(), updateTicker *time.Ticker) {
```
antonblock, line 538: Running `make generate` deletes this, it should be in a different file

internal/graphql/schema.resolvers.go
```diff
@@ -33,10 +33,15 @@ import (
...
 	"github.com/observiq/bindplane-op/internal/util/semver"
 	"github.com/observiq/bindplane-op/model"
+	"github.com/observiq/bindplane-op/model/graph"
 	"github.com/observiq/bindplane-op/model/otel"
 	"go.uber.org/zap"
 )
 
+const configurationMetricsUpdateInterval = 10
+const overviewMetricsUpdateInterval = 10
+const agentMetricsUpdateInterval = 10
```
antonblock, line 43: These are also deleted by `make generate`

PR #399, opened 2022-09-26
feat: Support Processors in Destinations

andykellr:
See configuration-macos-destination-processors.yaml for an example of processors in a destination.

Resolves observiq/iris#1194

### Proposed Change
Renders destination processors before destinations. This also means that the processors will be included before any processors that are specified as part of the destination.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


jmwilliams89 (approved):

model/configuration.go
```diff
@@ -273,7 +273,23 @@ func evalDestination(destination *ResourceConfiguration, defaultName string, sto
 		return "", nil
 	}
 
-	return dest.Name(), destType.eval(dest, errorHandler)
+	destName := dest.Name()
+	partials := destType.eval(dest, errorHandler)
+
+	destProcessors := otel.NewPartials()
+	// evaluate the processors associated with the source
```
andykellr, line 280:
```suggestion
	// evaluate the processors associated with the destination
```

model/otel/configuration.go
```diff
@@ -178,22 +178,54 @@ func (p *Partial) Size() int {
...
 
+func prepend[T any](list []T, others ...T) []T {
+	var result []T
+	result = append(result, others...)
+	result = append(result, list...)
+	return result
+}
+
+// Prepend adds components from another partial by prepending each of the component lists together. Note: The
+// implementation may cause the Partial to be modified.
```
andykellr, line 197:
```suggestion
// Prepend adds components from another partial by prepending each of the component lists together.
```

antonblock (approved):

shazlehu (approved):

PR #406, opened 2022-09-26
Processor type documentation

jsirianni:
### Proposed Change

Documentation for the current set of processor types. 

**Note:** The directory path used for the processor docs may change depending on how readme.io requires docs to be structured when syncing from the repo to readme. I will finalize this before merging. If you have a suggestion for a doc structure, please post it here.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


ryancgoins:
Two thoughts: 

1. I think we should include a screenshot of the example as it would be configured in the UI so that the UI use is clear as well
2. It might just be me, but I frequently find myself confused by the difference between a record attribute and a resource attribute and I'm never sure which should be used. So some explanation of that in those processors I think would be helpful

Otherwise these look great

docs/resources/processor-types/add_attribute.md
```diff
@@ -0,0 +1,78 @@
...
+
+| Metrics | Logs | Traces |
+| :--- | :--- | :--- |
+|  | ✓ |  |
+
+## Configuration Table
+
+| Parameter  | Type    | Default  | Description |
+| :---       | :---    | :---     | :--- |
+| action     | `enum`  | `upsert` | insert: Add attribute if it does not exist. update: Update existing value. upsert: Insert or update. |
```
antonblock, line 16:
```suggestion
| action     | `enum`  | `upsert` | `insert`: Add attribute if it does not exist. `update`: Update existing value. `upsert`: Insert or update. |
```

jsirianni:
> Two thoughts:
> 
>     1. I think we should include a screenshot of the example as it would be configured in the UI so that the UI use is clear as well
> 
>     2. It might just be me, but I frequently find myself confused by the difference between a record attribute and a resource attribute and I'm never sure which should be used. So some explanation of that in those processors I think would be helpful
> 
> 
> Otherwise these look great

I think understanding the Open Telemetry data model is important but processor docs may not be the best place for it. A higher level doc would be useful.

I embedded some image links to processor configuration screenshots.

ryancgoins (approved):

Mrod1598 (approved):

PR #408, opened 2022-09-26
fix: Skip config prompt when upgrading debian package with apt-get

jsirianni:
### Proposed Change

When upgrading from v1.0.1 to v1.1.0, apt will prompt the user to replace or retain their current configuration. This is because of https://github.com/observIQ/bindplane-op/pull/374.

This causes upgrades via the install script to fail because the script is unattended, and the user cannot respond to the prompt.

We want to avoid replacing a user's configuration with the default configuration, therefor we should skip the prompt. I attempted the same upgrade using yum (centos 7) and dnf (centos 9 and fedora 35) and did not run into this issue.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


antonblock (approved):
Confirmed on Linux that 1.0.1 -> 1.1.0 upgrade didn't prompt me to change my config, didn't modify the existing config, and completed successfully

andykellr (approved):

PR #429, opened 2022-10-04
fix(session): Don't overwrite lookup error with save error

antonblock:
### Proposed Change
If we get an error looking up the session, preserve it so we can fall back on it after saving the session

##### Checklist
- [ ] Changes are tested
- [x] CI has passed


Mrod1598 (approved):

jsirianni (approved):

internal/server/auth/session.go
```diff
@@ -33,9 +33,9 @@ func CheckSession(server server.BindPlane) gin.HandlerFunc {
...
 			session.Options.MaxAge = 0
 
-			err := session.Save(c.Request, c.Writer)
-			if err != nil {
-				c.AbortWithError(http.StatusInternalServerError, err)
+			saveErr := session.Save(c.Request, c.Writer)
+			if saveErr != nil {
+				c.AbortWithError(http.StatusInternalServerError, saveErr)
 				return
 			}
```
dehaansa, line 41: We still want to abort on https://github.com/observIQ/bindplane-op/pull/429/files#diff-921a6c9586a3d9009ed0dc3d50e9a6cea57040e85cb922c0517c5f4426911aacR42 even when we were able to successfully save the session?

antonblock: I think we do because we still got an error retrieving the cookie https://github.com/observIQ/bindplane-op/pull/156

PR #452, opened 2022-10-11
Fix(UI): Do not show "Recent Telemetry" button on agent page that use Raw configurations.

dsvanlani:
Resolves: https://github.com/observIQ/iris/issues/1193

### Proposed Change
Current behavior: User navigates to an agent page.  The agent is using a raw config and the "Recent Telemetry" button is displayed.  User clicks the button and the dialog is empty.  The dialog will never populate because there is no "snapshot" processors in the config, and we haven't communicated this.

Proposed change: Hide this button for agents using Raw configurations.

While it's technically possible for a user to author a config which uses snapshot processors appropriately and can then use the feature it doesn't seem likely.  Let's hide it for now.

image

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


andykellr (approved):

ui/src/pages/agents/agent.tsx
```diff
@@ -126,7 +126,26 @@ const AgentPageContent: React.FC = () => {
...
+      return null;
+    }
+
+    return data.configurations.configurations.find(
+      (c) => c.metadata.name === configName
+    );
+  }, [data?.agent, data?.configurations]);
+
   const viewTelemetryButton = useMemo(() => {
+    if (currentConfig?.spec.raw !== "") {
```
dehaansa, line 145: Should this show a reason like the other disable reasons, rather than just hiding the button?

dsvanlani: We discussed this - but theres really no documentation on how a user could author a raw config to use the snapshot feature so we're hiding it for now.

dehaansa (approved):

PR #453, opened 2022-10-11
Fix: Some longer source names aren't centered in InlineSourceCards.

dsvanlani:
### Proposed Change
Apache Common and other two line source names looked bad in the current `InlineSourceCard`.

We can get around this by setting the height to 120px on the card and the "action area" of the card.

Before:
image

After:
image

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


jsirianni (approved):
Looks good, had this issue on `main`.

shazlehu (approved):

PR #462, opened 2022-10-12
Additional tests for measurements, and allow lookback if current bucket appears incomplete

dehaansa:
### Proposed Change
Fast follow for topology to add more testing and one more edge case of jumping back a bucket if most recent bucket has less data than the immediately previous bucket (only tested once), so we hopefully don't show users weird changes in data.

##### Checklist
- [X] Changes are tested
- [x] CI has passed


dehaansa:
Failing test is the inconsistent throughput test andy's working on where the measurements are only added to a source in one pipeline, not all.

dsvanlani (approved):

andykellr (approved):

PR #464, opened 2022-10-12
feat: Disable sources and destinations

antonblock:
### Proposed Change
Support "pausing" sources and destinations using a new flag `Disabled`. If a source or destination has that set, during `Render()` it will be omitted from the final pipeline

- Refactor `ResourceConfiguration` to just have `Name` and `ParameterizedSpec`
- Update tests
- Add tests for rendering configs with disabled source or destinations

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


model/configuration_test.go
```diff
@@ -985,3 +1141,20 @@ func TestDuplicate(t *testing.T) {
 		require.Equal(t, new.Spec.Selector.MatchLabels["configuration"], duplicateName)
 	})
 }
+
+func TestConfigurationType(t *testing.T) {
+	t.Run("raw configuration", func(t *testing.T) {
+		c := NewConfigurationWithSpec("raw config", ConfigurationSpec{
```
dsvanlani, line 1147: This absolutely does not matter but technically `raw config` isn't a valid configuration name.  Could change it to `raw-config` or leave it.

dsvanlani (approved):
Looks good to me.  I PRd #466 into this branch because I thought it was clearer to use another source type in the `configuration-macos-source-disabled.yaml` besides `macos` but I think the current test is fine.  Feel free to include it or merge as is 👍

Mrod1598 (approved):

PR #469, opened 2022-10-13
feat: Custom sources and destinations

shazlehu:
### Proposed Change
Adds custom sources & destinations, where the user can insert their own yaml into a config. Resolves observIQ/iris#1181

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


andykellr (approved):

resources/source-types/custom.yaml
```diff
@@ -0,0 +1,46 @@
+apiVersion: bindplane.observiq.com/v1
+kind: SourceType
+metadata:
+  name: custom
+  displayName: Custom
+  description: Insert a custom OpenTelemetry receiver configuration.
+  icon: /icons/destinations/custom.svg
```
jsirianni, line 7: This works but it is not using the source custom.svg.

jsirianni: We could share a single svg for both, or fix this path.

jsirianni (changes requested):

jsirianni (approved):

PR #470, opened 2022-10-14
fix(UI): Save + Edit destinations fixed on the Components + Overview page.

dsvanlani:
### Proposed Change
- Add an `enableProcessors` prop to the `ResourceDestinationCard`, and pass it inside the configuration page but default it to false.
- Only attempt to save processors on a configuration when the `destinationIndex` is known and the updated processors are not empty.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


antonblock (approved):
Edited a destination from the overview page and saw the config apply successfully

Mrod1598 (approved):
Edited a destination and it worked. LGTM.

PR #472, opened 2022-10-15
feat(UI): Default to first active telemetry in PipelineGraph and OverviewGraph

dsvanlani:
### Proposed Change
- `model.Graph` now has a top level attributes of type `map[string]any`
- We are including the `activeTypeFlags` in this top level attribute for the overview graph and configuration graph.
- OverviewGraph sets the selected telemetry to the first active telemetry type after load.
- PipelineGraph determines the first selected telemetry from the configuration graph prop.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


shazlehu (approved):
As Dave & I discussed, I don't love the useMemo, but otherwise seems sound, and functions well on my machine.

andykellr (approved):

PR #490, opened 2022-10-17
fix: Always include destination d0 metrics

andykellr:
### Proposed Change
If d0 metrics aren't measured, rates will be confusing after processors are added. A follow-on PR for observiq/iris#1217 will optimize this with a single throughputmeasurement processor.

This also includes a fix where destinations with multiple sources are not adding those sources together.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


andykellr:
Closing in favor of #493

PR #491, opened 2022-10-17
feat(destinationtype): Splunk HTTP Event Collector (HEC)

jsirianni:
### Proposed Change

Added Splunk HTTP Event Collector destination type. The exporter allows OTLP data be translated to Splunk HEC format and sent to an HTTP Event Collector, which can then be ingested into the Splunk logging platform. Note that this solution is different from Splunk Observability Cloud, which is already supported by BindPlane OP and uses a different set of exporters. Splunk HEC supports logs only.

**Configuration**

Screenshot from 2022-10-17 15-48-27

Screenshot from 2022-10-17 15-48-53

**Testing**

Tested with on premise Splunk and vCenter log plugin.

Screen Shot 2022-10-17 at 3 25 11 PM
Screen Shot 2022-10-17 at 3 24 50 PM

##### Checklist
- [x] Changes are tested
- [x] CI has passed


dsvanlani (approved):
Form looks good and I verified that I was able to generate a valid configuration with this for my agent. 👍

antonblock (approved):
Form looks good, config sent to my agent was correct

PR #493, opened 2022-10-18
feat: Optimize measurement processors when no processors exist

andykellr:
### Proposed Change
If there are no processors and two adjacent s0/s1 or d0/d1 processors would exist, use a single "s" or "d" processor. On ingest, expand them to s0/s1 and d0/d1 to keep metrics consistent when processors are added.

Includes changes from #490 which I will close

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


model/otel/configuration.go
```diff
@@ -493,7 +493,7 @@ func (p *Pipeline) AddReceivers(rc *RenderContext, id []ComponentID) {
 func (p *Pipeline) AddProcessors(rc *RenderContext, ids []ComponentID) {
 	// when adding processors, only add measurement processors once
```
dehaansa, line 494: Should this comment be updated?

model/configuration_test.go
```diff
@@ -884,7 +1014,7 @@ service:
                 - resourceattributetransposer/source0__processor0
                 - resourceattributetransposer/source0__processor1
                 - throughputmeasurement/_s1_metrics_source0
-                - throughputmeasurement/_d1_metrics_googlecloud
+                - throughputmeasurement/_d_metrics_googlecloud
```
dehaansa, line 1017: Is there a good way to optimize this? Since it's still two throughput measurement processors in sequence?

andykellr: Yes

dsvanlani (approved):

model/configuration_test.go
andykellr: After discussing with Sam we are going to table this further optimization as the throughputmeasurement processor may be replaced soon.

dehaansa (approved):

PR #502, opened 2022-10-24
fix(source-type): Remove newlines from credentials field

dsvanlani:
### Proposed Change
When using the `multiline` option for the `credentials` field, the editor preserves newlines.  Previously newlines would just be removed and the credentials would be set on one line.  When using the `mulitline` field this added newlines to the template which caused the `yaml` that was rendered as an otel config to be invalid.

A simple workaround to get the `multiline` input is to make sure we remove newlines from this field when rendering.

Example:

A destination configured as follows:

image

Can now properly render it's configuration.

#### Caveats

We'll need to be conscious of this behavior when using the `multiline` option.  Though this is the only instance where we use it. 

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


jsirianni (approved):
Working great for me :+1:

Mrod1598 (approved):
Worked for me as well

PR #507, opened 2022-10-26
added readme automation workflow

asenkowski:
### Proposed Change
 Added Readme.io sync to github workflow

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


jsirianni (approved):
Looks good.

Readme automation development was done using the command used here. I will be rotating the secret `README_API_KEY` with a fresh API key for the readme site.

.github/workflows/release.yml
```diff
@@ -65,3 +65,22 @@ jobs:
...
+    runs-on: "ubuntu-20.04"
+    needs:
+      - release
+    steps:
+      - name: Check out source code
+        uses: actions/checkout@v3
+
+      - name: "Get Tag From Environment"
+        id: get-tag
+        run: printf '::set-output name=tag::%s' "$(printf '%s' "${{ github.ref }}" | sed 's/refs\/tags\///')"
```
antonblock, line 79: We're getting warnings about `set-output` being deprecated. Since this is new, it'd be great to use the officially supported method. Details in their [blog post](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)

jsirianni: We paired on this and pushed a change. The new solution should work by taking the tag `v1.4.0` and stripping off the `v` prefix, followed up by setting VERSION in the GITHUB_ENV file.

antonblock (approved):

PR #509, opened 2022-10-27
feat: MongoDB Atlas source

cpheps:
### Proposed Change
Created a MongoDB Atlas source.

Some limitations:
- The receiver allows multiple `project` structures to be specified for alerts and logs. Sources doesn't currently support an array of structs so I made it so you can only configure a single project. You would need multiple instances of this source for multiple projects.
- The alert collection in `poll` mode ideally needs a storage extension. There's currently no way to reference a storage extension in the `receiver` block.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


resources/source-types/mongodb_atlas.yaml
```diff
@@ -0,0 +1,507 @@
...
+      description: How often (seconds) to scrape for metrics.
+      type: int
+      default: 60
+      relevantIf:
+        - name: enable_metrics
+          operator: equals
+          value: true
+      advancedConfig: true
+
+    - name: granularity
```
jsirianni, line 50: I think it would be nice to link to documentation here.

Example from Elasticsearch
```yaml
    - name: nodes
      label: Collection Scope
      documentation:
        - text: Read more
          url: https://www.elastic.co/guide/en/elasticsearch/reference/7.9/cluster.html#cluster-nodes.
```

jsirianni: Also, this option does not seem to be used in the receiver config.

jsirianni (changes requested):

jsirianni:
We have been putting source docs at [docs/www/sources](https://github.com/observIQ/bindplane-op/tree/main/docs/www/sources). This may be out of scope for you, I think we should at least track it if we do not implement a doc here. Docs will be published to readme.io during a release.

jsirianni (commented):
I was able to get metrics / logs working with ease :+1:

resources/source-types/mongodb_atlas.yaml
cpheps: Oh cool I didn't know I could do that.

cpheps:
>We have been putting source docs at [docs/www/sources](https://github.com/observIQ/bindplane-op/tree/main/docs/www/sources). This may be out of scope for you, I think we should at least track it if we do not implement a doc here. Docs will be published to readme.io during a release.

I'm not sure to be honest I didn't know we had docs. How are docs handled for sources that have so much conditional configuration? Do we just list them out best we can?

jsirianni:
> 

Yeah, there is not a specification on how it is handled right now. A table with each option and an example on how to use the source in its most default form is probably fine.

jsirianni (approved):

docs/www/sources/mongodb-atlas.md
```diff
@@ -0,0 +1,44 @@
+---
+title: "MongDB Atlas"
+category: 633dd7654359a20031653089
+slug: "mogondb0atlas"
```
antonblock, line 4:
```suggestion
slug: "mongodbatlas"
```
Not sure if that's intentional for the slug

cpheps: nope fat fingered it.

cpheps: @antonblock fixed

antonblock (approved):

PR #520, opened 2022-11-01
fix(UI): Do not provide "Delete Destination" option on Overview Page

dsvanlani:
### Proposed Change
There's really no concept of deleting a destination in this case because:
1. Destinations are associated with configurations and can only be removed by editing the config.
2. While we can delete a Destination resource, this isn't allowed because this destination is in use by configurations.

Config page has delete option on destinations:
 
image

Overview page does not:

image

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


shazlehu (approved):
Looks good , works well on my machine. 

Nit: All the linter changes make it hard to see what your actual changes are. Wonder if we could try splitting them out into a separate commit in the future? Or doing the lint in a batch.

armstrmi (commented):
LGTM, runs great over here as well. One question tho, I'm having trouble seeing exactly where the delete button is removed. Not terribly well versed in React, so I was hoping that you could enlighten exactly where that happens

armstrmi (approved):

Mrod1598 (approved):

PR #535, opened 2022-11-08
feat(parameters): Add NotEquals operator to relevantIf

antonblock:
### Proposed Change
Add `notEquals` as an operatort for `RelevantIf`. It's the opposite of `equals`, so parameters will be shown when the other parameter value doesn't match

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


ui/src/components/ResourceConfigForm/satisfiesRelevantIf.ts
```diff
@@ -12,10 +12,16 @@ export function satisfiesRelevantIf(
   const relevantIf = definition.relevantIf;
 
   for (const condition of relevantIf) {
-    // TODO (dsvanlani) Right now we only support and expect the "equals" operator
-    // Add a capability to satisfy other operators like "less than" or "greater than".
-    if (!isEqual(formValues[condition.name], condition.value)) {
-      return false;
+    if (condition.operator === RelevantIfOperatorType.Equals) {
```
dsvanlani, line 15: This works great, I cant tell if this would be more readable:

```suggestion
   for (const condition of relevantIf) {
    switch (condition.operator) {
      case RelevantIfOperatorType.Equals:
        if (!isEqual(formValues[condition.name], condition.value)) {
          return false;
        }
        break;
      case RelevantIfOperatorType.NotEquals:
        if (isEqual(formValues[condition.name], condition.value)) {
          return false;
        }
        break;
    }
  }
```

dsvanlani (approved):

andykellr (approved):

PR #537, opened 2022-11-08
fix(plugin): Remove duplicate timezone parameter from common event format

jsirianni:
### Proposed Change

The `location` parameter is not used, and the source type already has a `timezone` parameter that serves the same purpose.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


dehaansa (approved):

dsvanlani (approved):
LGTM.  I was worried that this would break editing an existing config with CEV source, but I verified that it works.

PR #555, opened 2022-11-18
fix(bindplanectl): Disable resty warning for HTTP

antonblock:
### Proposed Change
Disable the warning resty logs if using HTTP. It makes it seem like a command failed, but is fine since the server was configured to use HTTP.

Before:
```console
nico@pc> bindplanectl get configurations
2022/11/18 11:42:06.418093 WARN RESTY Using Basic Auth in HTTP mode is not secure, use HTTPS
NAME                   	MATCH
Docker                 	configuration=Docker
```

After:
```console
nico@pc> bindplanectl get configurations
NAME                   	MATCH
Docker                 	configuration=Docker
```

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


andykellr (approved):
Good call, should have done this long ago.

jsirianni (approved):

jsirianni:
We had some internal discussion about warning the user once, when using http. I think it would be nice to put this warning into the init step. If the user is not enabling / configuring TLS during setup, we could provide a warning there.

PR #559, opened 2022-11-21
feat(ui): Overview graph destinations navigate to new Destinations tab.

shazlehu:
### Proposed Change

The destination nodes in the overview graph navigate to new Destinations tab, which replaces the Components tab.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


ui/src/components/Tables/ComponentsTable/ComponentsDataGrid.tsx
```diff
@@ -9,7 +9,7 @@ import {
 } from "@mui/x-data-grid";
```
dsvanlani, line 9: We can probably rename this file to `DestinationsDataGrid.tsx`

dsvanlani (changes requested):
Works great.  I think we can remove all references to "Components" here so we don't confuse anyone later.  Which would mean:

1. Rename ComponentsTable and ComponentsDataGrid and test files.
2. Change the query to `Destinations` instead of `Components`
3. `resourcesFromSelected` function can be simpler, we don't need to parse the Type out of the grid selection model.  Just need to return something in the form:
```
{
    kind: "Destination",
    metadata: { name },

**Review by antonblock — COMMENTED** [review]
Looks good aside from Dave's comments

File: ui/src/components/Cards/OverviewDestinationCard.tsx [hunk 2]
```diff
@@ -0,0 +1,104 @@
...
+  CardContent,
+  Stack,
+  Typography,
+} from "@mui/material";
+import { useSnackbar } from "notistack";
+import { memo } from "react";
+import { classes } from "../../utils/styles";
+
+import styles from "./cards.module.scss";
+import { useNavigate } from "react-router-dom";
```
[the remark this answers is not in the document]
armstrmi: Is there a reason for moving those two to the upper block? Just Curious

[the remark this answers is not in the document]
dsvanlani: Not really other than I try and keep the style imports in a separate block.  Just preference.

ui/src/components/Cards/OverviewDestinationCard.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+interface ResourceDestinationCardProps {
+  name: string;
+  // disabled indicates that the card is not active and should be greyed out
+  disabled?: boolean;
+  enableProcessors?: boolean;
+}
+
+const OverviewDestinationCardComponent: React.FC<ResourceDestinationCardProps> =
+  ({ name, disabled, enableProcessors = false }) => {
```
antonblock, line 24:
```suggestion
  ({ name, disabled }) => {
```

ui/src/components/Cards/OverviewDestinationCard.tsx
```diff
@@ -0,0 +1,103 @@
...
+import { useNavigate } from "react-router-dom";
+import { useGetDestinationWithTypeQuery } from "../../graphql/generated";
+
+import styles from "./cards.module.scss";
+
+interface ResourceDestinationCardProps {
+  name: string;
+  // disabled indicates that the card is not active and should be greyed out
+  disabled?: boolean;
+  enableProcessors?: boolean;
```
antonblock, line 20:
```suggestion
```

antonblock (commented):
We should remove `enableProcessors` if it's never used

antonblock (approved):

ui/src/pages/components/DestinationsPage.tsx
```diff
@@ -1,32 +1,26 @@
 import { gql } from "@apollo/client";
```
dsvanlani, line 1: Sorry, but can we rename this directory from `components` to `destinations`

dsvanlani (approved):
One more directory name change requested.  After that looks good.

PR #571, opened 2022-11-29
feat(source): OracleDB Metrics source

antonblock:
### Proposed Change

- Update `Oracle Database` source to optionally collect metrics. This is done using the `oracledb_metrics` plugin in the collector.
- Add `enable_logs` flag for existing log parameters
- Bump the minor version of the source type since it's adding new functionality
- Update the documentation with new parameters and instructions on setting up the LPU

### Testing

If you're testing this, message me for access to a live system

##### Checklist
- [x] Changes are tested
- [x] CI has passed


docs/www/integrations/sources/oracle-database.md
```diff
@@ -9,20 +9,57 @@ hidden: false
...
+| username*           | `string`  |                                                                            | Database user to run metric queries with.                                      |
+| password            | `string`  |                                                                            | Password for user.                                                             |
+| sid                 | `string`  |                                                                            | Site Identifier. One or both of sid or service_name must be specified.         |
+| service_name        | `string`  |                                                                            | OracleDB Service Name. One or both of sid or service_name must be specified.   |
+| wallet              | `string`  |                                                                            | OracleDB Wallet file location (must be URL encoded).                           |
+| collection_interval | `int`     |                                                                            | How often (seconds) to scrape for metrics.                                     |
```
armstrmi, line 55: should list default value of 60

armstrmi (commented):
Can't currently approve this, awaiting on collector fix for plugin receiver.

resources/source-types/oracledb.yaml
```diff
@@ -4,20 +4,107 @@ metadata:
...
+
+    - name: endpoint
+      label: Endpoint
+      description: Address to scrape metrics from, as hostname:port.
+      type: string
+      default: localhost:1521
+      relevantIf:
+        - name: enable_metrics
+          operator: equals
+          value: true
```
jsirianni, line 29: Can we split this up into two options? I think the user experience is better, and the port can be defaulted to 1521.  Docs would need a change as well.

resources/source-types/oracledb.yaml
```diff
@@ -4,20 +4,107 @@ metadata:
...
+      label: Username
+      description: Database user to run metric queries with.
+      type: string
+      required: true
+      relevantIf:
+        - name: enable_metrics
+          operator: equals
+          value: true
+
+    - name: password
```
jsirianni, line 41: Is password required?

resources/source-types/oracledb.yaml
```diff
@@ -115,9 +224,41 @@ spec:
...
+
+  metrics:
+    receivers: |
+      {{ if .enable_metrics }}
+      - plugin/oracledb_metrics:
+          path: $OIQ_OTEL_COLLECTOR_HOME/plugins/oracledb_metrics.yaml
+          parameters:
+            endpoint: {{ .endpoint }}
+            username: {{ .username }}
+            password: {{ .password }}
```
jsirianni, line 245: We might want to single quote this to avoid weird issues with characters 
```suggestion
            password: '{{ .password }}'
```

resources/source-types/oracledb.yaml
```diff
@@ -4,20 +4,107 @@ metadata:
...
+
+    - name: service_name
+      label: Service Name
+      description: OracleDB Service Name. One or both of sid or service_name must be specified.
+      type: string
+      default: ""
+      relevantIf:
+        - name: enable_metrics
+          operator: equals
+          value: true
```
jsirianni, line 69: I think it would be nice to wrap these with an enum multi select. We could have the user choose their "connection type"
- sid
- service name
- sid and service name

This might allow us to set sid and service name with `required: true`.

antonblock: Thinking about how to do this, I like combining these to have validation, but need to figure out how multi-select would work

jsirianni: I cannot remember the details, but I thought we implemented multi select enum to solve something similar. @dsvanlani  might know more.

jsirianni (approved):

PR #579, opened 2022-12-05
feat: Prometheus Remote Write destination

cpheps:
### Proposed Change
Added a destination for Prometheus Remote Write exporter. Tested it against a local Cortex instance using [these](https://cortexmetrics.io/docs/getting-started/) instructions.

Screen Shot 2022-12-05 at 1 25 00 PM

Proof data is coming into Cortex:

Screen Shot 2022-12-05 at 1 24 20 PM

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


docs/www/integrations/destinations/prometheus-remote-write.md
```diff
@@ -0,0 +1,33 @@
...
+
+| Metrics | Logs | Traces |
+| :------ | :--- | :----- |
+| ✓       |      |        |
+
+## Configuration Table
+
+| Parameter                               | Type     | Default | Description                                                                                                                                                                           |
+| :-------------------------------------- | :------- | :------ | :------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| endpoint\*                              | `string` |         | Remote write URL to send remote write samples.                                                                                                                                        |
```
jsirianni, line 19: Can we break this up into a couple options?
- host: required
- port: default prometheus port

Protocol could be set conditionally, based on the `enable_tls` option.

resources/destination-types/prometheus_remote_write.yaml
```diff
@@ -0,0 +1,177 @@
...
+      type: string
+      required: false
+      default: ""
+      relevantIf:
+        - name: enable_tls
+          operator: equals
+          value: true
+      advancedConfig: true
+
+    - name: cert_file
```
jsirianni, line 50: Nit: It might be easier for the user to reason about if we have an `enable_mutual_tls` option, which then exposes the cert and key options. This way they can enable TLS without seeing the mutual tls options. I suspect most users will not be using mutual tls.

docs/www/integrations/destinations/prometheus-remote-write.md
jsirianni: Path would also need to be an option, and could probably have a default value and hidden under `advanced`.

resources/destination-types/prometheus_remote_write.yaml
cpheps: I don't disagree. I just copied the current way we're doing this. I think I copied it directly from another destination. Do we currently have a destination that does this? If so wondering if we should make a pass at some point to make them all the same.

docs/www/integrations/destinations/prometheus-remote-write.md
```diff
@@ -0,0 +1,33 @@
...
+| headers                                 | `map`    |         | Additional headers to attach to each HTTP Request. The following headers cannot be change: `Content-Encoding`, `Content-Type`, `X-Prometheus-Remote-Write-Version`, and `User-Agent`. |
```
andykellr, line 20:
```suggestion
| headers                                 | `map`    |         | Additional headers to attach to each HTTP Request. The following headers cannot be changed: `Content-Encoding`, `Content-Type`, `X-Prometheus-Remote-Write-Version`, and `User-Agent`. |
```

resources/destination-types/prometheus_remote_write.yaml
```diff
@@ -0,0 +1,177 @@
...
+          value: true
+      advancedConfig: true
+
+    # Advanced Options
+    - name: headers
+      label: Addition HTTP Headers
+      description: |
+        Additional headers to attach to each HTTP Request.
+
+        The following headers cannot be change: Content-Encoding, Content-Type, X-Prometheus-Remote-Write-Version, User-Agent
```
andykellr, line 80:
```suggestion
        The following headers cannot be changed: Content-Encoding, Content-Type, X-Prometheus-Remote-Write-Version, User-Agent
```

docs/www/integrations/destinations/prometheus-remote-write.md
cpheps: No we can't as it's an actual API endpoint depending on the tech. So for cortex it's `http://localhost:9009/api/v1/push`. The tech defines the endpoint so there's no standard format.

resources/destination-types/prometheus_remote_write.yaml
jsirianni: There is inconsistency right now for sure. Elasticsearch does this https://github.com/observIQ/bindplane-op/blob/main/resources/source-types/elasticsearch.yaml#L145

cpheps: Cool I'll change over to using Elasticsearch's model if we prefer that.

docs/www/integrations/destinations/prometheus-remote-write.md
cpheps: I mean I guess we can but I don't see an advantage to breaking this field up into three separate fields. I'm not sure if there is a default port for prometheus remote write. Looking now, if there is this makes sense if not then I don't see a huge advantage.

cpheps: Sorry slowly realizing what you were getting at. Where `enable_tls` sets the protocol. Yeah looks like port is different for every tech and so is path. So the user would have to specify everything we just auto populate the protocol I guess.

[Here's](https://prometheus.io/docs/operating/integrations/) the list of supported integrations. I randomly picked some and they all look unique.

andykellr:
@jsirianni If you're able to confirm that this works, I am happy with the configurations it produces with the various options I tried.

jsirianni (commented):
When setting only the required params (hostname, port, api path), I get the following error:
```
{"level":"error","timestamp":"2022-12-05T15:55:03.160-0500","message":"1 error occurred:\n\t* template: prometheus_remote_write:8:22: executing \"prometheus_remote_write\" at <.headers>: map has no entry for key \"headers\"\n\n","status":500,"method":"GET","path":"/v1/configurations/re","query":"","ip":"::1","user-agent":"go-resty/2.7.0 (https://github.com/go-resty/resty)","latency":0.005573429}
```

Some template issue with headers

PR #585, opened 2022-12-08
feat: Edit Source + Destination Processors in modal.

dsvanlani:
The PipelineGraph Processor Nodes are not interactive, showing a Dialog with the information about processors with the ability to add, reorder, edit, and delete.

image

### Proposed Change
- Add GQL Mutations for `addProcessor`, `updateProcessors`, and `updateProcessor`.
- Remove the Processors section from the normal resource configuration form, now called `ResourceConfigurationView`

### Screen Shots

image

image

image

image

##### Checklist
- [ ] Changes are tested
- [x] CI has passed


internal/graphql/schema.resolvers.go
```diff
@@ -152,6 +152,88 @@ func (r *metadataResolver) Labels(ctx context.Context, obj *model.Metadata) (map
...
+		return nil, fmt.Errorf("invalid resource type, should be source or destination")
+	}
+
+	_, err = r.bindplane.Store().ApplyResources(ctx, []model.Resource{config})
+	return nil, err
+}
+
+// AddProcessor is the resolver for the addProcessor field.
+func (r *mutationResolver) AddProcessor(ctx context.Context, input model1.AddProcessorInput) (*bool, error) {
+	fmt.Println("here")
```
Mrod1598, line 210:
```suggestion
```

Mrod1598 (approved):
Overall looks good but does feel a little weird that the only way to add processors now is in the topology, where as there's two spots to add sources and destinations. also there's no way to look at the raw and be able to modify processors.

shazlehu (approved):

PR #604, opened 2022-12-13
fix(destinations): Pass `paused` prop for DestinationsDataGrid

antonblock:
### Proposed Change

Update `EditDestinationDialog` to set `paused` based on the retrieved destination, was previously defaulting to false

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


dsvanlani (approved):
Working for me now 👍

PR #631, opened 2022-12-14
fix(destination): Use processors from config when toggling pause

antonblock:
### Proposed Change
In `onTogglePause`, use destination processors from the configuration instead of getting them from the updated destination. The `BPDestination` object won't have processors on it

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


dsvanlani (approved):
Working for me

PR #633, opened 2022-12-15
fix: Provide version as global js variable

dsvanlani:
### Proposed Change
It was a known issue where the Bindplane Version (shown in the bottom left-hand corner) would read `unknown` despite a version being set.  This is because an unauthenticated request was cached by the browser.

This PR serves a generated js template at `/js/_globals.js` which provides the version on the `globals` object.

This can be used for enterprise improvements where server config can be exposed by the server without first making a REST request.

## Testing

In dev environments the version will always be "unknown".  Version is injected at build time.  To run a build

build the binary:
```sh
make build
```

run the server:
```
dist/bindplane_darwin_amd64/bindplane serve
```

Navigate to `http://localhost:3001`
##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


dehaansa (approved):

Mrod1598 (approved):
image
LGTM

PR #636, opened 2022-12-19
Map parameter key/value labels

andykellr:
### Proposed Change
I did this during observathon and think is worth keeping around. This allows the heading at the top of the map parameter columns to be specified.

This is the parameter definition I used:

```yaml
    - name: intervals
      label: Intervals
      type: map
      required: true
      options:
        gridColumns: 12
        labels:
          key: Threshold
          value: Reporting Interval
```

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


dsvanlani (approved):
LGTM.  Tested it out:

image

Mrod1598 (approved):
image
LGTM

PR #641, opened 2022-12-30
feat: Add bpComponentID template function to reference unique component IDs

andykellr:
### Proposed Change
When rendering ParameterizedType templates such as the source defined in `sourcetype-filelog-storage.yaml` in this PR, it can be necessary to reference the generated name of the component. A new go-template function `bpComponentID` provides access to the `ComponentIDProvider` instance used during template rendering.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


model/testfiles/sourcetype-filelog-storage.yaml
```diff
@@ -0,0 +1,122 @@
...
+      advancedConfig: true
+
+  logs:
+    extensions: |
+      - file_storage:
+          directory: /tmp/offset_storage_dir
+
+    receivers: |
+      - plugin:
+          path: $OIQ_OTEL_COLLECTOR_HOME/plugins/file_logs.yaml
```
schmikei, line 96: If this is using the `file_logs` plugin, then the name of this parameter would be `offset_storage_dir` https://github.com/observIQ/observiq-otel-collector/blob/main/plugins/file_logs.yaml

I am currently refactoring the source to move away from the plugin and use the filelog receiver which would hopefully reduce the amount of layers of renaming. Not sure for this test file if you just want to rename this parameter so that it is valid with the current plugin or refactor this to just use the [`filelogreceiver`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/filelogreceiver#configuration)

schmikei: After thinking about it some more, it doesn't really matter as these tests are just testing the render of the YAML, going to resolve

schmikei (approved):
This worked great for me doing some refactoring of the file_logs source! Would be good to get a more bindplane-op familiar person to approve as well though.

<details>
<summary> Screenshots </summary>

image

image

</details>

antonblock (approved):

PR #642, opened 2022-12-30
feat: switch File Log source to use `filelogreceiver` rather than plugin

schmikei:
It is desired to reduce abstraction and give users more capability when using the [`filelogreceiver`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/filelogreceiver#configuration)

### Proposed Change

Also added the following parameters:

- include_file_name_attribute
- include_file_path_attribute
- include_file_name_resolved_attribute
- include_file_path_resolved_attribute
- poll_interval
- max_concurrent_files
- offset_storage_dir
- multiline_line_end_pattern

This change relies on #641 

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


docs/www/integrations/sources/filelog.md
```diff
@@ -9,21 +9,29 @@ hidden: false
...
+| include_file_path_attribute  | `bool`    | false                            | Whether to add the file path as the attribute `log.file.path`.                                                                                                          |
+| include_file_name_resolved   | `bool`    | false                            | Whether to add the file name after symlinks resolution as the attribute `log.file.name_resolved`.                                                                       |
+| include_file_path_resolved   | `bool`    | false                            | Whether to add the file path after symlinks resolution as the attribute `log.file.path_resolved`.                                                                       |
+| offset_storage_dir           | `string`  | $OIQ_OTEL_COLLECTOR_HOME/storage | The directory that the offset storage file will be created. It is okay if multiple receivers use the same directory.                                                    |
```
jsirianni, line 32: It might be wroth mentioning that the observIQ OTEL Collector has $OIQ_OTEL_COLLECTOR_HOME set by default in its runtime environment, which will differ between platforms but generally the user does not need to worry about it.

I think it is okay to keep the description brief in the parameter definition that is shown in the UI.

resources/source-types/file_log.yaml
```diff
@@ -87,28 +143,46 @@ spec:
...
+          include_file_name_resolved: {{ .include_file_name_resolved_attribute }}
+          include_file_path: {{ .include_file_path_attribute }}
+          include_file_path_resolved: {{ .include_file_path_resolved_attribute }}
+          start_at: {{ .start_at }}
+          poll_interval: {{ .poll_interval }}ms
+          storage: {{ bpComponentID "file_storage" }}
+          operators:
+            {{ if eq .parse_format "json" }}
+            - type: json_parser
+              parse_to: body
```
jsirianni, line 176: This might have been discussed, so forgive me if I am re-opening a discussion. It might be nice to allow the user to decide where fields are parsed_to. I know we prefer body by default but the spec seems to imply attributes. It would be nice if the user could decide.

The same would go for Regex, if you choose to pursue this.

resources/source-types/file_log.yaml
```diff
@@ -87,28 +143,46 @@ spec:
...
-              {{ range $fp := .exclude_file_path }}
-              - '{{ $fp }}'
-              {{end}}
+            {{ range $fp := .exclude_file_path }}
+            - '{{ $fp }}'
+            {{end}}
+          {{ end }}
+          multiline:
+            line_start_pattern: '{{ .multiline_line_start_pattern }}'
+            line_end_pattern: '{{ .multiline_line_end_pattern }}'
```
jsirianni, line 164: I think when handling multiline you need to select one of the two available options (matching the beginning or end).

It might be nice to wrap this in an enum "selector" which allows the user to select "no multiline (default)", "match end", "match beginning".

jsirianni (commented):
This was not introduced here but we might want to fix it here. When selecting regex as a parse option, we might want to make regex_pattern required.

Also, I get this error when selecting regex without a regex pattern. I am not sure if that is the cause of this issue though:
```json
{"level":"error","timestamp":"2023-01-04T13:49:34.827-0500","message":"1 error occurred:\n\t* yaml: line 3: did not find expected '-' indicator\n\n","status":500,"method":"GET","path":"/v1/configurations/file","query":"","ip":"::1","user-agent":"go-resty/2.7.0 (https://github.com/go-resty/resty)","latency":0.003696033}
```

resources/source-types/file_log.yaml
schmikei: Ah  thats a good idea, I'll go ahead and make them exclusive 👍

resources/source-types/file_log.yaml
schmikei: Yeah I think my current understanding was that the old file log source directly in BPOP was supposed to go to body anyhow, however I wasn't sure if we wanted to expose potentially an enum for where to `parse_to` in order to do this.

@cpheps do you have any thoughts/preferences given the whole parsing to `body` and `attributes` still appears to be a fragile construct?

resources/source-types/file_log.yaml
schmikei: Made the change! ✅

docs/www/integrations/sources/filelog.md
schmikei: Added a blurb in the docs ✅

schmikei:
> This was not introduced here but we might want to fix it here. When selecting regex as a parse option, we might want to make regex_pattern required.

I made it required if parse format is regex now ✅

resources/source-types/file_log.yaml
cpheps: Ok looking at this. I think we should expose `parse_to` as an advanced field. It would be nice to choose this depending on your destination as they don't all handling parsing to body well.

schmikei: I added the extra parameter as an advanced configuration option 👍  

image

jsirianni (approved):

jmwilliams89 (approved):

PR #655, opened 2023-01-09
Fix: Discarding Changes after editing processor causes white screen

dsvanlani:
#### Bug FIx

Previously these steps would cause a Type error when discarding changes in the ProcessorDialog.

1. Add a new processor
2. Save it
3. Edit the newly added processor
4. Hit "X" or escape key, discard changes.

I can't explain why this only happened after editing a newly added processor and not editing any processor, but the editing processor index can certainly be null and we should handle that.

#### Save -> Done

Feedback offered in Demo that we should be saying "Done" here and not save.

image

image

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


shazlehu (approved):

Mrod1598 (approved):
Bug seems to be fixed for me.

PR #656, opened 2023-01-09
feat(processor): Add Mask processor

armstrmi:
Signed-off-by: armstrmi <mitchell.armstrong@bluemedora.com>

### Proposed Change

Added PII Processor that wraps [maskprocessor](https://github.com/observIQ/observiq-otel-collector/tree/main/processor/maskprocessor)

##### Checklist
- [x] Changes are tested
- [ ] CI has passed
Data:
```
{
  "email": "mitchell.armstrong@bluemedora.com",
  "email": "abc123._%+-|@abc123.-.hello",
  "email": "hello",
  "phone": "1-616-123-4567",
  "phone": "987978478977987897987",
  "ssn": "123-45-6789",
  "ssn": "123456789",
  "credit_card": "1234-1234-1234-1234",
  "credit_card": "111111111111111",
  "credit_card": "2222222222222222"
}

```
Result:
image


resources/processor-types/filter_pii.yaml
```diff
@@ -0,0 +1,68 @@
+apiVersion: bindplane.observiq.com/v1
+kind: ProcessorType
+metadata:
+  name: filter_pii
+  displayName: PII
```
shazlehu, line 5: Was this displayName vetted? If not, I think something more descriptive would be better, e.g. Mask Personal Identifiying Information.

resources/processor-types/filter_pii.yaml
```diff
@@ -0,0 +1,68 @@
...
+  description: Wraps mask processor to remove PII from log records
+spec:
+  version: 0.0.1
+  parameters:
+    - name: rules
+      label: Match Expression
+      description: Expr expression to find matching logs.
```
shazlehu, line 12: There should be a link the Expr language reference here, see "Count Logs" processor

resources/processor-types/filter_pii.yaml
antonblock: +1 to this. I think `Mask Sensitive Data` is a good label

resources/processor-types/filter_pii.yaml
```diff
@@ -0,0 +1,71 @@
+apiVersion: bindplane.observiq.com/v1
+kind: ProcessorType
+metadata:
+  name: filter_pii
+  displayName: Mask Personal Identifiying Information.
+  description: Wraps mask processor to remove PII from log records
```
antonblock, line 6:
```suggestion
  description: Detect and mask sensitive data in log records.
```

antonblock (changes requested):

resources/processor-types/filter_pii.yaml
jmwilliams89: I second Nico's suggestion. This processor can be used to mask any type of data. It doesn't have to be personally identifying. So sensitive data is probably more encompassing.

jmwilliams89: I'd suggesting changing the displayName and name to reflect this. This is also why the underlying processor is called `maskprocessor` because it doesn't have to be PII.

resources/processor-types/filter_pii.yaml
jmwilliams89: This actually isn't using the Expr language. Rules are a series of regex expressions to match on.

resources/processor-types/filter_pii.yaml
```diff
@@ -0,0 +1,71 @@
...
+      label: Match Expression
+      description: Expr expression to find matching logs.
+      type: map
+      documentation:
+        - text: How to Mask Attributes
+          url: https://github.com/observIQ/observiq-otel-collector/tree/main/processor/maskprocessor#how-it-works
+      options:
+        gridColumns: 12
+      default:
+        email: \b[a-z0-9._%\+\-—|]+@[a-z0-9.\-—|]+\.[a-z|]{2,6}\b
```
shazlehu, line 20: '—' character (registers as \U+FFE2) (between the escaped dash and the pipe) looks suspicious to me... Also, I'm pretty sure modern email has a more expansive character set.... Not sure how up to date we want this regexp to be

shazlehu: Also only masks email in lowercase

armstrmi: I agree with lowercase, probably worth while to include those, but I think for the most part we can exclude some of the more fancy characters typically included in spam type emails [!?$%....etc]. However, I do think it would be worthwhile to at least include the uppercase in the regex.

resources/processor-types/filter_pii.yaml
antonblock: @armstrmi Should update the description with that ☝️

resources/processor-types/filter_pii.yaml
```diff
@@ -0,0 +1,71 @@
+apiVersion: bindplane.observiq.com/v1
+kind: ProcessorType
+metadata:
+  name: sensitive_data_processor
+  displayName: Sensitive Data Masking.
+  description: Detect and mask sensitive data in log records.
+spec:
+  version: 0.0.1
+  parameters:
+    - name: rules
```
antonblock, line 10: This should be required, and then you won't need to check if it's set in the template

resources/processor-types/filter_pii.yaml
```diff
@@ -0,0 +1,66 @@
+apiVersion: bindplane.observiq.com/v1
+kind: ProcessorType
+metadata:
+  name: sensitive_data_processor
+  displayName: Sensitive Data Masking.
+  description: Detect and mask sensitive data in log records.
+spec:
+  version: 0.0.1
+  parameters:
```
antonblock, line 9: Probably also worth adding toggles for enabling on Logs/Metrics/Traces. Check out the custom source type for an example

resources/processor-types/filter_pii.yaml
```diff
@@ -31,6 +39,7 @@ spec:
 
   logs:
     processors: |
+      {{ if has "Logs" .telemetry_types }}
```
dehaansa, line 42: It feels cleaner to put this logic around the whole `logs:` section, to not have an empty `logs: processors: |` when logs is not requested. Same for other telemetry types

resources/processor-types/mask_sensitive_data.yaml
```diff
@@ -0,0 +1,79 @@
+apiVersion: bindplane.observiq.com/v1
+kind: ProcessorType
+metadata:
+  name: mask_sensitive_data
+  displayName: Mask Sensitive Data
+  description: Detect and mask sensitive data in log records.
```
antonblock, line 6:
```suggestion
  description: Detect and mask sensitive data in collected telemetry.
```

resources/processor-types/mask_sensitive_data.yaml
```diff
@@ -0,0 +1,79 @@
...
+spec:
+  version: 0.0.1
+  parameters:
+    - name: telemetry_types
+      label: Telemetry Types
+      type: enums
+      description: "Select which types of telemetry the processor supports."
```
antonblock, line 13:
```suggestion
      description: "Select which types of telemetry to apply masking rules to."
```

antonblock (approved):
Small nitpicks in descriptions, otherwise working well

PR #669, opened 2023-01-17
feat(topology): Show all telemetry on config topology

antonblock:
### Proposed Change

- Update `PipelineGraph` to show all telemetry types at once
- Add alternate `updateMetricData` function to fetch and format that

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


ui/src/components/PipelineGraph/Nodes/CustomEdge.tsx
```diff
@@ -1,15 +1,17 @@
-import { EdgeProps, getBezierPath } from "react-flow-renderer";
-import { classes } from '../../../utils/styles';
+import { EdgeProps, getBezierPath } from "reactflow";
+import { classes } from "../../../utils/styles";
 import { isNodeDisabled } from "./nodeUtils";
 
+// default styling
+import 'reactflow/dist/style.css';
```
andykellr, line 6: Noticed the quote style changed above from single to double and now this is single. Should we add this to the linter?

antonblock: Good catch. Yeah the presets we use don't seem to care, I'm in favor of enforcing double quotes

andykellr (approved):
Working well and the code looks clean. One question about single vs double quotes.

Screenshot 2023-01-24 at 2 28 12 PM

Also, I don't love the metric label format (too bold?, remove colons?) but I assume this matches the design.

andykellr (changes requested):
Labels always in center, processor spacing inconsistent.

ui/src/components/PipelineGraph/Nodes/CustomEdge.tsx
dsvanlani: +1

dsvanlani (commented):
I think we should dim telemetry labels as part of the hover functionality:

image

ui/src/components/MeasurementControlBar/MeasurementControlBar.tsx
```diff
@@ -28,22 +28,33 @@ export const TELEMETRY_SIZE_METRICS: { [telemetryType: string]: string } = {
...
   onPeriodChange: (period: string) => void;
+  onTelemetryTypeChange?: (telemetry: string) => void;
 }
 
 export const MeasurementControlBar: React.FC<MeasurementControlBarProps> = ({
   onTelemetryTypeChange,
   onPeriodChange,
   telemetry,
   period,
+  telemetryTypes = TELEMETRY_TYPES,
```
dsvanlani, line 52: I'd prefer to not set a default here, and just conditionally render the telemetry types if present.  

This makes it so you don't have to pass an empty object `{ }` into the MeasurementControlBar, just not passing telemetry types is sufficient.  Then in the OverviewGraph we can more explicitly pass in the telemetry types.  

I

ui/src/components/PipelineGraph/Edges/ConfigurationEdge.tsx
```diff
@@ -0,0 +1,87 @@
+import { Stack } from "@mui/material";
```
dsvanlani, line 1: Theres two files called ConfigurationEdge now.  Can we rename one? Probably this one to `CustomEdge.tsx` ?

ui/src/components/PipelineGraph/Edges/ConfigurationEdge.tsx
```diff
@@ -0,0 +1,87 @@
...
+        className="react-flow__edge-path"
+        d={edgePath}
+        markerEnd={markerEnd}
+        startOffset={startOffset}
+        textAnchor={textAnchor}
+      />
+      <EdgeLabelRenderer>
+        <div
+          style={{
+            position: "absolute",
```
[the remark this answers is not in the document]
dsvanlani: Ahh, I see we have to set the translate programmatically.

I'm okay with this for now.

dsvanlani (changes requested):
This is looking good.  It looks like the labels should be dimmed on hover but I didn't look into too much why they were not dimming.  It looks like we have access to that hovered set so I think it would be good to get in.

dsvanlani (approved):

ui/src/utils/graph/utils.ts
```diff
@@ -1,25 +1,48 @@
...
+  SourceAfterLabels, //s1
+  SourceAfterProcessors, // s2
+  LabelsAfterProcessors, // s3
+  DestinationBeforeLabels, // d0
+  DestinationAfterLabels, // d1
+  DestinationAfterProcessors, // d2
   Configuration,
 }
 
+/**
```
shazlehu, line 33: Gold star for finally commenting this beast!

shazlehu (approved):
Looks really nice, and again, thanks for adding sorely needed comments.

PR #683, opened 2023-01-23
feat(destination-type): Prometheus Remote Write destination

antonblock:
### Proposed Change

This is just https://github.com/observIQ/bindplane-op/pull/579 being readded now that the issue with seeding resources with map params is resolved

* Finalized prometheus remote write destionation
* Added documentation for Prom remote write
* Update docs/www/integrations/destinations/prometheus-remote-write.md
* Update resources/destination-types/prometheus_remote_write.yaml
* Updated TLs parameters and broke up URL
* Made some PR fixups

##### Checklist
- [x] Changes are tested
- [x] CI has passed


jsirianni:
Do we have control over positioning? HTTP header and external label options are not aligned. 


Screenshot from 2023-01-23 10-34-59

resources/destination-types/prometheus_remote_write.yaml
```diff
@@ -0,0 +1,217 @@
...
+      advancedConfig: true
+
+    - name: namespace
+      label: Namespace
+      description: Prefix to attach to each metric name.
+      type: string
+      default: ""
+      advancedConfig: true
+
+    - name: enable_resource_to_telemetry_conversion
```
jsirianni, line 128: I think the UI would look nicer if this option was above enable TLS

antonblock:
> Do we have control over positioning? HTTP header and external label options are not aligned.
> 

We don't have that level of control, but I can try to rearrange it

resources/destination-types/prometheus_remote_write.yaml
antonblock: @jsirianni updated the arrangement here

cpheps (approved):

resources/destination-types/prometheus_remote_write.yaml
```diff
@@ -0,0 +1,221 @@
...
+            {{ end }}
+          {{ end }}
+
+          {{ if ne .namespace "" }}
+          namespace: {{ .namespace }}
+          {{ end }}
+
+          {{ if .enable_write_ahead_log }}
+          wal:
+            directory: $OIQ_OTEL_COLLECTOR_HOME/storage/prom_rw
```
jsirianni, line 200: Will this cause an issue if we have multiple remote write destinations on a single system?

I configured two, and I see the following directory structure:

```
[root@fedora]/opt/observiq-otel-collector# tree storage 
storage
└── prom_rw
    └── prom_remotewrite
        └── 00000000000000000001
```

cpheps: So it appends the `prom_remotewrite` directory to whatever path is passed in [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/prometheusremotewriteexporter/wal.go#L90). We likely want make the `directory` parameter unique to each instance.

jsirianni (approved):

PR #708, opened 2023-01-31
Configure retry and queue for measurement exporter

jsirianni:
### Proposed Change

During an outage, collectors will retry sending measurements. While this is ideal, the measurement volume being sent per collector is too high, and can be difficult to recover from.

Set `retry_on_failure` to default values except for:
- `max_elapsed_time` is now 30 seconds instead of five minutes. After 30 seconds, the attempted measurement is dropped.
- `max_interval` is now 5s instead of 30s. This is to avoid backing off due to the shorter max time

Set `sending_queue` to default values except for:
- queue_size is now 60 instead of 5,000

You can find the types and default values for each configuration option here:
- Retry https://github.com/open-telemetry/opentelemetry-collector/blob/v0.70.0/exporter/exporterhelper/queued_retry.go#L249
- Queue https://github.com/open-telemetry/opentelemetry-collector/blob/v0.70.0/exporter/exporterhelper/queued_retry.go#L44

We have a ticket for making these options configurable by the user. This PR is strictly to serve as a starting point.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


jsirianni:
My collector config contains the following, when given a config by this build:

```yaml
exporters:
    otlphttp/_agent_metrics:
        endpoint: http://fedora:3001/v1/otlphttp
        retry_on_failure:
            enabled: true
            initial_interval: 5s
            max_elapsed_time: 30s
            max_interval: 5s
        sending_queue:
            enabled: true
            num_consumers: 10
            queue_size: 60
```

antonblock (approved):

andykellr (approved):

PR #710, opened 2023-02-02
feat(agent): Support removing configurations

antonblock:
### Proposed Change

- Add a button to the `Choose another configuration` menu labeled `Remove Configuration`. It will remove the `Configuration` label, making that agent unmanaged
- Add more detailed tracing to that API endpoint
- Add tests

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


dsvanlani (changes requested):
1. We should not show the "Remove Configuration" button when the agent is unmanaged.
2. The page looks slightly frenetic when we click this because of the editing state changing before we get the new agent.  I think ideally we could create a `mutation` that patches the labels and returns the agent, then when we receive that agent we can update the cache immediately so there is no flash of a previous config.

dsvanlani (approved):

internal/graphql/schema.resolvers.go
```diff
@@ -189,6 +192,33 @@ func (r *mutationResolver) UpdateProcessors(ctx context.Context, input model1.Up
...
+	agent, err := r.bindplane.Store().Agent(ctx, input.AgentID)
+	if err != nil {
+		span.SetStatus(codes.Error, err.Error())
+		return nil, err
+	}
+
+	newAgent, err := r.bindplane.Store().UpsertAgent(ctx, agent.ID, func(current *model.Agent) {
+		newLabels := current.Labels
+		newLabels.Set["configuration"] = ""
+		current.Labels = newLabels
```
BinaryFissionGames, line 211: Nit: This can be reduced to `current.Labels.Set["configuration"] = ""`, since maps are pointers in go.

PR #719, opened 2023-02-07
fix(sourcetype): Remove page_in and page_out metrics from Host source type

jsirianni:
### Proposed Change

I noticed an issue when configuring `page_in` and `page_out` metrics from the process group in the Host source type. These two metrics are not in the [metric list](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/hostmetricsreceiver/internal/scraper/pagingscraper/metadata.yaml).

Screenshot from 2023-02-07 13-03-56

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


Mrod1598 (approved):
Yeah, looks like they were moved into operations with labels

dsvanlani (approved):

PR #737, opened 2023-02-14
Fix: remove mui icons dep

shazlehu:
### Proposed Change
While looking at this dependably pull (https://github.com/observIQ/bindplane-op/pull/735), realized that we only use one icon from the library in one place, so moved it to our repo's icons, removed the dependency. 

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


dsvanlani (approved):

antonblock (approved):
Nice catch

PR #739, opened 2023-02-15
doc: Clarify which platforms are supported for bindplane server

jsirianni:
### Proposed Change

Currently Linux is the only officially supported solution for BindPlane server. This change re-words some confusion text in the installation doc to clarify that Linux is the only platform supporting BindPlane server.

##### Checklist
- [x] Changes are tested
- [x] CI has passed


docs/www/installation.md
```diff
@@ -6,7 +6,10 @@ hidden: false
 ---
 ## Install BindPlane OP
 
-BindPlane OP does not have any dependencies and can run on Windows, Linux, or macOS although we recommend installing BindPlane OP server on a supported Linux distribution.
+BindPlane OP has two components, server and and optional command line interface.
```
dsvanlani, line 9:
```suggestion
BindPlane OP has two components, server and an optional command line interface.
```

cpheps (approved):
We generate Windows zip files and executables. Should we not do so if we don't plan to support them?

dsvanlani (approved):

jsirianni:
> We generate Windows zip files and executables. Should we not do so if we don't plan to support them?

Not sure. I think we do intend to support Windows at some point. We are just missing the MSI and script to make it easy to install and run as a service.

cpheps:
> Not sure. I think we do intend to support Windows at some point. We are just missing the MSI and script to make it easy to install and run as a service.

Yeah, I think we should have CI jobs that build and test on Windows but I don't think we should have Windows deliverables if we explicitly say we don't support them. Seems like a mixed message. Probably the same for macOS if we concerned with supporting that for customer.

jsirianni:
> > Not sure. I think we do intend to support Windows at some point. We are just missing the MSI and script to make it easy to install and run as a service.
> 
> Yeah, I think we should have CI jobs that build and test on Windows but I don't think we should have Windows deliverables if we explicitly say we don't support them. Seems like a mixed message. Probably the same for macOS if we concerned with supporting that for customer.

I agree. I passed this onto @ryancgoins. I think we either remove the builds, or we start supporting windows soon.

PR #743, opened 2023-02-16
feat: Refactored store updates

jmwilliams89:
### Proposed Change
The following PR will refactor store updates and invert control. This is necessary for a future PR that will restructure the store package. The underlying behavior of transitive updates should remain the same.

### Improvements
- Decouples `Updates` and `Store`. At the moment, there is a circular dependency between the two. `Store` creates an instance of `Updates` when applying resources. It then passes itself in as a parameter to `Updates`. This change will remove that circular dependency. The `Store` struct will have full query control and determine how to add transitive updates after applying a resource. The `Updates` struct will be treated as purely a data structure, with no control.
- Comments and names have been added or modified to increase readability. Previously, it was difficult to infer the intent of `Updates` or `storeUpdates`. This PR should help increase that readability.
- There was previously a possibility of a nil pointer exception when calling any of the include functions on `Updates`. These functions have been updated to instantiate the underlying data structure if it doesn't exist, before then adding to it.
- The previous design queried for all configurations no matter what when adding transitive updates. This means that an update to an agent or agent version would have trigged a query for all configurations, even though this would have no effect on them. The new design will only query for configurations if there's a possibility configurations were affected.
- The new design has 100% test coverage and should be easier to unit test as changes are made.

##### Checklist
- [x] Changes are tested
- [x] CI has passed


internal/store/boltstore.go
```diff
@@ -281,10 +281,51 @@ func (s *boltstore) ApplyResources(ctx context.Context, resources []model.Resour
 	return resourceStatuses, errs
 }
 
+// addTransitiveUpdates adds all the transitive updates based on the resource type.
+func (s *boltstore) addTransitiveUpdates(ctx context.Context, updates *Updates) error {
```
cpheps, line 285: `updates` function parameter shadows the struct field `updates `*UpdateEventBus`

internal/store/googlecloudstore.go
```diff
@@ -540,11 +540,52 @@ func (s *googleCloudStore) UserSessions() sessions.Store {
 // ----------------------------------------------------------------------
 // events
 
+// addTransitiveUpdates adds all the transitive updates based on the resource type.
+func (s *googleCloudStore) addTransitiveUpdates(ctx context.Context, updates *Updates) error {
```
cpheps, line 544: I didn't see an update to changes the `updates` struct field. It looks like it's different than other store types? 

https://github.com/observIQ/bindplane-op/blob/tech-debt/store-updates/internal/store/googlecloudstore.go#L47

internal/store/mapstore.go
```diff
@@ -675,8 +675,49 @@ func (mapstore *mapStore) upsertAgent(agentID string, updater AgentUpdater, upda
 	return agent
 }
 
+// addTransitiveUpdates adds all the transitive updates based on the resource type.
+func (mapstore *mapStore) addTransitiveUpdates(ctx context.Context, updates *Updates) error {
```
cpheps, line 679: Same as bbolt `updates` shadows the struct field.

cpheps (approved):
LGTM few nitpicks

internal/store/boltstore.go
jmwilliams89: I don't think it does shadow it actually. If you wanted the UpdatesEventBus, you would call `s.updates` instead of `updates` in this context. I do think the event bus struct field should be renamed, because it's confusing, but I was going to save that for a future PR. I didn't want to expand the scope of this any further.

cpheps: Yup...you're right. I had two train of thought going when I wrote the comment.

antonblock (approved):

PR #745, opened 2023-02-16
feat: Added Log Deduplication processor

cpheps:
### Proposed Change
Adds the log deduplication processor type

Screenshot 2023-02-21 at 9 59 33 AM

##### Checklist
- [X] Changes are tested
- [ ] CI has passed


jsirianni:
I think it would be nice if we could specify a field to dedup against.

The bindplane healtcheck log would be a great candidate for this processor but it does not work. I am assuming that is because the `latency` is going to differ.

```json
{"level":"info","timestamp":"2023-02-16T17:27:33.999-0500","message":"/health","status":200,"method":"GET","path":"/health","query":"","ip":"127.0.0.1","user-agent":"curl/7.82.0","latency":0.00001065}
{"level":"info","timestamp":"2023-02-16T17:27:34.204-0500","message":"/health","status":200,"method":"GET","path":"/health","query":"","ip":"127.0.0.1","user-agent":"curl/7.82.0","latency":0.000014027}
{"level":"info","timestamp":"2023-02-16T17:27:34.410-0500","message":"/health","status":200,"method":"GET","path":"/health","query":"","ip":"127.0.0.1","user-agent":"curl/7.82.0","latency":0.000009347}
{"level":"info","timestamp":"2023-02-16T17:27:34.615-0500","message":"/health","status":200,"method":"GET","path":"/health","query":"","ip":"127.0.0.1","user-agent":"curl/7.82.0","latency":0.000009868}
{"level":"info","timestamp":"2023-02-16T17:27:34.821-0500","message":"/health","status":200,"method":"GET","path":"/health","query":"","ip":"127.0.0.1","user-agent":"curl/7.82.0","latency":0.000014527}
{"level":"info","timestamp":"2023-02-16T17:27:35.026-0500","message":"/health","status":200,"method":"GET","path":"/health","query":"","ip":"127.0.0.1","user-agent":"curl/7.82.0","latency":0.000013756}
{"level":"info","timestamp":"2023-02-16T17:27:35.232-0500","message":"/health","status":200,"method":"GET","path":"/health","query":"","ip":"127.0.0.1","user-agent":"curl/7.82.0","latency":0.000009107}
```

This would mean a change to the underlying processor, so I think we can move forward with this as is.

jsirianni (approved):

resources/processor-types/log_deduplicate.yaml
```diff
@@ -0,0 +1,37 @@
...
+      type: string
+      required: true
+      default: UTC
+
+  logs:
+    processors: |
+      - logdedup:
+          interval: {{ .interval }}s
+          timezone: {{ .timezone }}
+          log_count_attribute: {{ .log_count_attribute }}
```
BinaryFissionGames, line 37: This is something Dave noticed with my PR, we should protect against users inserting quotes and malforming the yaml
```suggestion
          timezone: {{ .timezone | quote }}
          log_count_attribute: {{ .log_count_attribute | quote }}
```

cpheps: Cool I didn't realize this was a thing.

cpheps:
@jsirianni Yeah I think we're already seeing a strong case for an `exclude_fields` parameter or something to make the matching not so strict. I'll follow up with product and see if we want to make that change before the next release.

cpheps:
Going to put this on hold as we're adding some parameters to the underlying processor to address matching criteria.

cpheps:
@BinaryFissionGames @jsirianni Can I get a re-review of this. Updated logdedup processor with `exclude_fields`.

resources/processor-types/log_dedup.yaml
```diff
@@ -0,0 +1,51 @@
+apiVersion: bindplane.observiq.com/v1
+kind: ProcessorType
+metadata:
+  name: log_dedup
+  displayName: DeDuplicate Logs
```
BinaryFissionGames, line 5: Thoughts on changing capitalization here? 
```suggestion
  displayName: Deduplicate Logs 
```

resources/processor-types/log_dedup.yaml
```diff
@@ -0,0 +1,51 @@
...
+      type: strings
+      default: []
+
+    - name: timezone
+      label: Timezone
+      advancedConfig: true
+      description: Timezone is used to localize the observed timestamp attributes.
+      type: string
+      required: true
+      default: UTC
```
BinaryFissionGames, line 38: Something I notice here - the "Advanced" section doesn't show up when creating the processor (this field is just present), but it does appear when editing an existing one. I don't think it should be addressed in this PR, just noting.

cpheps: We should file a bug against that behavior.

BinaryFissionGames: I've created a bug ticket in the backlog.

BinaryFissionGames (approved):

jmwilliams89 (approved):

PR #752, opened 2023-02-17
fix(measurements): Create unique receivers for each processor creating log-based metrics

antonblock:
### Proposed Change

In this context, "processors" is only referring to `logcount` and `metricextract`, which need to output to a route receiver.

- Currently we have a hardcoded route receiver called `builtin` which all processors needing it route to. This works and its measurements are reflected at destination processors, but we want to be able to see which processor is creating which measurements.
- Update `eval` so that each processor uses the last part of its`bpComponentID` as the name of the route receiver.
- When building the pipeline, insert route receivers with those names and add a throughput measurement processor to it.
- We'll then be able to use the `bpComponentID` in the front end to determine which processor created the telemetry

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


model/resource_type.go
```diff
@@ -163,6 +166,15 @@ func (rt *ResourceType) templateFuncComponentID(nameProvider otel.ComponentIDPro
 		return string(nameProvider.ComponentID(componentName)), nil
 	}
 }
+func (rt *ResourceType) templateFuncRouteID(nameProvider otel.ComponentIDProvider) func() (string, error) {
+	return func() (string, error) {
+		componentID, err := rt.templateFuncComponentID(nameProvider)("route")
+		if err != nil {
+			return "", err
+		}
+		return strings.ReplaceAll(componentID, "route/", ""), nil
```
cpheps, line 175: nit: Do we want to really replace all instances of `route/` or just the first? I'm not sure if there's a situation where you could have something like `route/route/`

cpheps (approved):
Tested in staging and works as expected.

dsvanlani (approved):
Working well for me

image

PR #761, opened 2023-02-20
fix(measurements): Show log-based metrics after source processor

antonblock:
### Proposed Change

In `updateMetricData`, assign metrics created by log-based metrics processors to that source's processor. So instead of only displaying at d0, also display at s1.

- Add tests
- Add comments

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


ui/src/utils/graph/utils.ts
```diff
@@ -608,11 +631,56 @@ export function updateMetricData(
...
+ */
+export function getMetricForNode(
+  nodeID: string,
+  metrics: GraphMetric[],
+  telemetryType: string
+): GraphMetric | undefined {
+  return metrics.find(
+    (m) =>
+      m.nodeID === nodeID && m.name === TELEMETRY_SIZE_METRICS[telemetryType]
+  ) ?? findRouteReceiverMetric(nodeID, metrics, telemetryType);
```
andykellr, line 649: Instead of only looking for the route receiver metric if there is no node metric, we should always look for it and add it in. That way if the source produces metrics and logs and we add a log count processor, the count metrics are added to the existing metrics produced by the source.

andykellr (changes requested):
It should add source metric throughput and processor metric throughput to produce a total.

Mrod1598 (approved):

dsvanlani (approved):

PR #767, opened 2023-02-21
fix(configuration): Remove raw view from config page

antonblock:
### Proposed Change

- On the config page, only show topology.
- Remove the switcher, remove the `PipelineGraph` from a Stack to fix spacing

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


ui/src/pages/configurations/configuration/index.tsx
```diff
@@ -190,19 +187,13 @@ const ConfigPageContent: React.FC = () => {
 
       {!isRaw && (
         <CardContainer>
-          <Stack spacing={2}>
-            <RawOrTopologyControl
-              rawOrTopology={rawOrTopology}
-              setTopologyOrRaw={setTopologyOrRaw}
-            />
```
BinaryFissionGames, line 197: I think we might want to replace this with a header saying `Topology` or something like that, so it's consistent with how the Agent's card container also has a title.

BinaryFissionGames (commented):
Does it make sense to remove the raw config from the GQL query, or does that break stuff?

antonblock:
> Does it make sense to remove the raw config from the GQL query, or does that break stuff?

It's kind of a misleading name, `raw` is the field for if someone manually does YAML instead of using our sources/destinations. We could get rid of `rendered` at the bottom, though, which is the YAML we're no longer showing

ui/src/pages/configurations/configuration/index.tsx
dsvanlani: I agree, the other cards have titles.

Not sure if Configuration or Topology is correct.  Product question.

dsvanlani (approved):
Code looks good, might want to ask product about a card Title.

ui/src/pages/configurations/configuration/index.tsx
antonblock: Updated with a centered "Topology" label, I'll run it by product

BinaryFissionGames (approved):

PR #768, opened 2023-02-22
fix(host_source): Fixed Optional Host metrics not being enabled

cpheps:
### Proposed Change
- Added `DefaultDisabled` field to `MetricOption` to indicate if it's disabled by default in OTel
- Updated the host source template so the optional metrics now enable correctly see example config

```yaml
receivers:
    hostmetrics/source0:
        collection_interval: 60s
        scrapers:
            cpu:
                metrics:
                    system.cpu.time:
                        enabled: false
                    system.cpu.utilization:
                        enabled: true
            filesystem:
                metrics:
                    system.filesystem.utilization:
                        enabled: true
            load:
                metrics: null
            memory:
                metrics:
                    system.memory.utilization:
                        enabled: true
            network:
                metrics:
                    system.network.conntrack.count:
                        enabled: true
                    system.network.conntrack.max:
                        enabled: true
            paging:
                metrics:
                    system.paging.utilization:
                        enabled: true
```

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


antonblock (approved):
Works as expected disabling/enabling both default and optional metrics

dsvanlani (approved):
Working as expected, good find.

PR #776, opened 2023-02-23
fix: Added generate command for server mocks and moved protocol

jmwilliams89:
### Proposed Change
- Added mock commands for `server.Manager`, `server.BindPlane`, and `server.Protocol`
- Moved `server.Protocol` into its own package. This prevents cyclical imports when testing with a protocol mock in the server package (see manager_test.go).
- Moved the contents of the `report` package into the `protocol` package. This gives the `protocol` package consistent ownership of all structs it sends to an agent.
- Renamed `report` structs to better reflect their intent and improve readability elsewhere in the codebase.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


antonblock (approved):

internal/server/protocol/report.go
```diff
@@ -12,8 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package report provides the Configuration for requesting a snapshot report from an agent.
-package report
+package protocol
```
dsvanlani, line 15: Is `snapshot` a better name here?

dsvanlani (approved):
LGTM.  Ignore the package comment about `snapshot`, I like `protocol`.

PR #777, opened 2023-02-23
fix: Added commands to generate client and measurement mocks

jmwilliams89:
### Proposed Change
- Added a generate command for client mocks
- Added a generate command for measurement mocks
- Fixed tests that use these mocks. Previously, the mocks were manually updated, which ended up subverting the mocking pattern and breaking tests with newly generated mocks.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


internal/cli/commands/get/testutil.go
```diff
@@ -37,20 +37,8 @@ var yamlOutput = "yaml"
...
-// Agents TODO(doc)
-func (c *mockClient) Agents(_ context.Context, _ ...client.QueryOption) ([]*model.Agent, error) {
-	return []*model.Agent{
-		{
-			ID:              "1",
+	agent1 := &model.Agent{
+		ID:              "1",
 			Architecture:    "amd64",
 			HostName:        "local",
 			Platform:        "linux",
```
BinaryFissionGames, line 44: this should be formatted with go fmt:
```
go fmt ./...
```

BinaryFissionGames (approved):

cpheps (approved):

PR #789, opened 2023-02-27
fix: Remove dummy processor interactions

Mrod1598:
### Proposed Change
Remove the hover interaction for the dummy processor node and change to default cursor.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


antonblock (approved):
LGTM, cursor changes when the processor node is editable, stays as a pointer when there are no sources and/or destinations

dsvanlani (approved):
LGTM

PR #795, opened 2023-03-01
feat: Grafana Loki DestinationType

dsvanlani:
### Proposed Change
Adds the Grafana Loki Destination type, using the `loki` exporter.

Misc:
- Add package `helpers` and `exporterHelper` to help standardize displaying `retry_on_failure` and `send_queue` options for exporters.
- Bool Param Inputs with `sectionHeader: true` now can display documentation.

#### Resource Helpers

Since the options for `exporterhelper` can be used by many exporters, I tried to start a pattern that would make it easy to consistently reuse the parameters.  We have slight inconsistencies in our `tls` configuration implementations that are hard to fix without breaking changes, so I want to get it right the first time.

`resources/helpers/exporter-helper/exporter_helper.go` contains `gotemplate` functions to render the otel config portion.  We then use them in the `FuncMap` being passed into the evaluate portion of rendering an otel config.

`resources/helpers/exporter-helpers/parameters.yaml` contain the standard parameters we should be using in DestinationTypes.  Its still a copy and paste, but at least its in a designated file and we don't have to copy and paste from another DestinationType.

### Screenshots

image

image

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


resources/destination-types/loki.yaml
```diff
@@ -0,0 +1,22 @@
...
+metadata:
+  name: loki
+  displayName: Grafana Loki
+  description: Send logs to Loki.
+  icon: /icons/destinations/loki.svg
+
+spec:
+  version: 0.0.1
+  parameters:
+    - name: endpoint
```
jmwilliams89, line 12: Can we add some of the advanced HTTP parameters as advanced settings to this destination: https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/confighttp/README.md

I think TLS and header support may be important for some users.

jsirianni (commented):
Loki is generally self hosted, and can be prone to performance issues if some of the sending queue / retry logic is not configured.

This is an example of what I have been using:

```yaml
loki:
  endpoint: http://loki:3100/loki/api/v1/push
  retry_on_failure:
    enabled: false
  sending_queue:
    enabled: true
    queue_size: 1000
```

If you decide to add these params, lets keep in mind that they apply to all exporters, so we will want to standardize on them here so we can easily copy them to other exporters in the future. https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/README.md

resources/destination-types/loki.yaml
```diff
@@ -0,0 +1,264 @@
...
+      required: false
+      default: ""
+      relevantIf:
+        - name: enable_tls
+          operator: equals
+          value: true
+        - name: mutual_tls
+          operator: equals
+          value: true
+      advancedConfig: true
```
jsirianni, line 106: Unclear to me if Loki supports mutual TLS or not. This blog indicates it does not:
```
Unfortunately, Loki doesn't support mutual TLS yet. But, everything that will connect to Loki does support mutual TLS.
```
https://smallstep.com/blog/grafana-mtls-homelab-observability/#loki

Grafana's docs also claim Loki does not support authentication of any kind, which probably includes mutual TLS.

jsirianni: thinking about this a bit further, I think we should keep this. It is possible Loki would be running behind a proxy which does have mutual tls enabled.

jsirianni (approved):

PR #825, opened 2023-03-07
feat(sourcetype): Add Sap Netweaver source type 

JonathanWamsley:
### Proposed Change

Adds SAP Netweaver source type based on the collector plugin.

SAP Netweaver Source in Add Source modal:
SAP Netweaver Source Type

SAP Netweaver Source Config:
SAP Netweaver Source Config

SAP Netweaver Source Config 2

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


docs/www/integrations/sources/sap-netweaver.md
```diff
@@ -0,0 +1,34 @@
+---
```
jsirianni, line 1: We are no longer contributing docs to the repo. Team will review this and then remove it before merging. We will also handle porting it into readme.io.

resources/source-types/sap_netweaver.yaml
```diff
@@ -0,0 +1,262 @@
...
+kind: SourceType
+metadata:
+  name: netweaver
+  displayName: SAP Netweaver
+  icon: /icons/sources/sap.svg
+  description: Collect metrics from SAP Netweaver.
+spec:
+  version: 0.0.1
+  supported_platforms:
+    - linux
```
jsirianni, line 11: Can the collector run remotely / from a different system than the SAP system? If so, I think we can support linux, windows, and macos here.

resources/source-types/sap_netweaver.yaml
```diff
@@ -0,0 +1,262 @@
...
+  parameters:
+    # Metrics
```
jsirianni, line 13: Because this is a metrics only source type (does not support logs / trace), we can remove the `enable_metrics` logic.

resources/source-types/sap_netweaver.yaml
```diff
@@ -0,0 +1,262 @@
...
+      default: ""
+      relevantIf:
+        - name: enable_metrics
+          operator: equals
+          value: true
+      advancedConfig: true
+
+    - name: profile
+      label: Profile
+      description: The profile path in the form of /sapmnt/SID/profile/SID_INSTANCE_HOSTNAME to collect sapnetweaver.abap.rfc.count and sapnetweaver.abap.session.count metrics.
```
jsirianni, line 77: Is this a file on the SAP system? If so, it probably means the collector must run on the SAP system directly? If so, ignore my first comment on supported platforms.

We still might support Windows, as SAP Netweaver does have a windows option AFAIK.

resources/source-types/sap_netweaver.yaml
```diff
@@ -0,0 +1,262 @@
...
+      type: string
+      default: ""
+      relevantIf:
+        - name: enable_metrics
+          operator: equals
+          value: true
+      options:
+        sectionHeader: true
+      advancedConfig: true
```
jsirianni, line 86: If profile is optional, I think we should mention what it enables when used, and possibly link to external documentation if you think that would be useful.

```suggestion
      advancedConfig: true
      documentation:
        - text: Read more
          url: https://linktodoc.com/somepath
```

resources/source-types/sap_netweaver.yaml
```diff
@@ -0,0 +1,262 @@
...
+
+
+  metrics:
+    receivers: |
+      {{ if .enable_metrics }}
+      - sapnetweaver:
+          collection_interval: {{ .collection_interval }}s
+          endpoint: {{ .hostname }}:{{ .port }}
+          username: {{ .username }}
+          password: {{ .password }}
```
jsirianni, line 243: Sometimes fields like this can behave funny if not wrapped in single quotes.

```suggestion
          username: '{{ .username }}'
          password: '{{ .password }}'
```

resources/source-types/sap_netweaver.yaml
```diff
@@ -0,0 +1,262 @@
...
+          {{ if .profile }}
+            profile: {{ .profile }}
```
jsirianni, line 245:
```suggestion
            profile: '{{ .profile }}'
```

jsirianni, line 245: No need to indent here
```suggestion
          {{ if .profile }}
          profile: {{ .profile }}
```

jsirianni (changes requested):

docs/www/integrations/sources/sap-netweaver.md
JonathanWamsley: removed

resources/source-types/sap_netweaver.yaml
JonathanWamsley: ignored as stated below

resources/source-types/sap_netweaver.yaml
JonathanWamsley: updated

resources/source-types/sap_netweaver.yaml
JonathanWamsley: I asked Project manager Joe, and he said to just support linux unless we hear otherwise

resources/source-types/sap_netweaver.yaml
JonathanWamsley: add docs. It is a little confusing because the [reference docs](https://help.sap.com/docs/SAP_NETWEAVER_750/b17bd20044a44edb9aa7830ee52ffac3/00fabccaa98348f291f3c7f02098116b.html) show the base path to be `/usr/sap/<SID>/SYS/profile` however when investigating and talking to SAP experts, we used `/sapmnt/SID/profile` which [are basically the same thing](https://answers.sap.com/answers/5487635/view.html) and is softlink there

resources/source-types/sap_netweaver.yaml
JonathanWamsley: updated

resources/source-types/sap_netweaver.yaml
JonathanWamsley: updated

JonathanWamsley: updated

resources/source-types/sap_netweaver.yaml
```diff
@@ -0,0 +1,225 @@
...
+      default: ""
+      advancedConfig: true
+
+    - name: password
+      label: Password
+      description: The password to use when connecting to SAP Netweaver.
+      type: string
+      required: true
+      default: ""
+      advancedConfig: true
```
jsirianni, line 47: Lets remove `advancedConfig` from these, as they are required options and should be front and center.

resources/source-types/sap_netweaver.yaml
```diff
@@ -0,0 +1,225 @@
...
+          operator: equals
+          value: true
+      advancedConfig: true
+
+
+  metrics:
+    receivers: |
+      - sapnetweaver:
+          collection_interval: {{ .collection_interval }}s
+          endpoint: {{ .hostname }}:{{ .port }}
```
jsirianni, line 205: I think this is missing the protocol. Lets do `https` if enable_tls is true, else `http`.

resources/source-types/sap_netweaver.yaml
```diff
@@ -0,0 +1,225 @@
...
+          username: '{{ .username }}'
+          password: '{{ .password }}'
+          {{ if .profile }}
+          profile: '{{ .profile }}'
+          {{ end }}
+          tls:
+          {{ if .enable_tls }}
+            insecure: false
```
jsirianni, line 213: We [recently discovered](https://github.com/observIQ/bindplane-op/pull/829) that tls.insecure is gRPC specific, and is not required if using the otel http client.

resources/source-types/sap_netweaver.yaml
```diff
@@ -0,0 +1,225 @@
...
+            insecure_skip_verify: {{ .insecure_skip_verify }}
+            ca_file: "{{ .ca_file }}"
+            cert_file: "{{ .cert_file }}"
+            key_file: "{{ .key_file }}"
+          {{ else }}
+            insecure: true
+          {{ end }}
```
jsirianni, line 220: See previous comment

```suggestion
          {{ end }}
```

jsirianni (changes requested):
I am having some issues due to the protocol not being injected into the `endpoint` parameter. See comments.

resources/source-types/sap_netweaver.yaml
JonathanWamsley: removed

resources/source-types/sap_netweaver.yaml
JonathanWamsley: added protocol but for .configure_tls instead

resources/source-types/sap_netweaver.yaml
JonathanWamsley: applied similar changes

resources/source-types/sap_netweaver.yaml
JonathanWamsley: updated

resources/source-types/sap_netweaver.yaml
jsirianni: I think `enable_tls` is fine, because it does indicate whether TLS is going to be used or not.

The Loki resource is different because we rely on the user to specify the protocol (Loki's endpoint parameter is a full URL).

The logic in the template looks good though, I just think we can revert back to enable_tls.

jsirianni (approved):

jsirianni:
Looking good

Screenshot from 2023-03-09 17-30-39

jmwilliams89 (approved):

jsirianni (approved):

PR #830, opened 2023-03-08
feat: Refactor store creation

jmwilliams89:
### Proposed Change
- Refactored configuration for stores
- Refactored store defaults
- Added a build function for stores
- Updated config conversion for stores
- Removed datastore/pubsub from config

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


configconvert/convert.go
```diff
@@ -119,6 +98,16 @@ func copyCommon(legacyCfg common.Common) config.Common {
 	return newCfg
 }
 
+func copyStoreConfig(legacyCfg common.Server) storeconfig.Config {
+	return storeconfig.Config{
```
cpheps, line 102: This is probably fine for oss but should we actually look at the StoreType to see what it is? We don't need to create a bbolt config if type isn't bbolt.

cpheps (approved):

store/config/config.go
```diff
@@ -0,0 +1,94 @@
...
+
+// BuiltInSecret is the secret used to encode sessions in the store.
+var BuiltInSecret = uuid.NewString()
+
+// Config is the configuration for a store.
+type Config struct {
+	// Type is the type of store.
+	Type string `mapstructure:"type,omitempty" yaml:"type,omitempty"`
+	// Secret is the secret used to encode sessions in the store.
+	Secret string `mapstructure:"secret,omitempty" yaml:"secret,omitempty"`
```
Mrod1598, line 47: I don't hate calling this just secret. However, when it's just under store it makes me think that it might be used for more than just session secrets. Also so now we have `secret` and `secretkey` even if it's under store, still feels like it could be confusing.

store/config/config.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/google/uuid"
+	"github.com/observiq/bindplane-op/store"
+	"go.uber.org/zap"
+)
+
+const (
+	// StoreTypeMap uses an in-memory store
+	StoreTypeMap = "map"
+	// StoreTypeBbolt uses go.etcd.io/bbolt for storage
+	StoreTypeBbolt = "bbolt"
```
antonblock, line 32: This should probably match the casing from `DefaultBBoltName` below, one way or the other

store/config/config.go
```diff
@@ -0,0 +1,94 @@
...
+
+	switch cfg.Type {
+	case StoreTypeMap:
+		return store.NewMapStore(ctx, options, logger), nil
+	case StoreTypeBbolt:
+		db, err := store.InitDB(cfg.BBolt.Path)
+		if err != nil {
+			return nil, fmt.Errorf("bbolt storage file failed to open: %w", err)
+		}
+		return store.NewBoltStore(context.Background(), db, options, logger), nil
```
antonblock, line 90:
```suggestion
		return store.NewBoltStore(ctx, db, options, logger), nil
```

store/config/config.go
```diff
@@ -0,0 +1,94 @@
...
+		Secret:    BuiltInSecret,
+		BBolt: BBolt{
+			Path: path.Join(bindplaneHome, DefaultBBoltName),
+		},
+	}
+}
+
+// Build will build a store from the configuration.
+func Build(ctx context.Context, cfg *Config, logger *zap.Logger) (store.Store, error) {
+	if cfg.Secret == "" {
```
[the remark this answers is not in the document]
jmwilliams89: Yeah. It's being generated in the global var `BuiltInSecret` when creating a default config. This means when initConfig is called and there isn't a secret in the config, the runtime will use its own.

internal/cli/commands/serve/start.go
```diff
@@ -67,19 +66,13 @@ func (s *Server) Start(ctx context.Context, bindplane *cli.BindPlane, h profile.
 		return err
 	}
 
-	// If session secret is not set, set it randomly at startup.
-	if config.SessionsSecret == "" {
-		config.SessionsSecret = uuid.NewString()
```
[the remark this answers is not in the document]
jmwilliams89: This is being handled in `BuiltInSecret` and when we init the config.

configconvert/convert_test.go
```diff
@@ -21,6 +21,7 @@ import (
 	"github.com/observiq/bindplane-op/config"
 	"github.com/observiq/bindplane-op/logging"
 	"github.com/observiq/bindplane-op/model/version"
+	storeconfig "github.com/observiq/bindplane-op/store/config"
```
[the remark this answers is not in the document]
jmwilliams89: They should be as we break up the overall config into more sections. The trace package is another good example of this.

store/config/config.go
jmwilliams89: I went with the more generic `secret` in case we need to perform any other encoding in the store. At the moment, it's used for just sessions, and perhaps we'll never need it for additional encoding. But this should make it flexible in the future.

Mrod1598 (approved):

antonblock (approved):

PR #838, opened 2023-03-09
feat: Generate secret key if it doesn't exist on startup

BinaryFissionGames:
### Proposed Change
* Disallow running insecure with no secret key
  * If a secret key is not specified in the config on startup, generate one and write it to the config
  * If no secret key is specified, and a config was not specified, fail to start (e.g. in containerized deployments)
* Remove now-unused profile.Helper parameter. 

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


BinaryFissionGames:
Looking into this build failure: https://github.com/observIQ/bindplane-op/actions/runs/4377317078/jobs/7660926787

The secret key isn't being set on the config when using the flag on docker, trying to debug why that's the case.

jsirianni:
Might need to update the tests here https://github.com/observIQ/bindplane-op/blob/gen-secret-key-if-not-exist/client/integration_test.go#L52, similar issue as inspec I bet.

jsirianni:
Tested this against docker and it works great.
```bash
➜  bindplane-op git:(gen-secret-key-if-not-exist) ✗ docker run -it observiq/bindplane-amd64
{"level":"error","timestamp":"2023-03-09T23:35:23.558Z","message":"unable to Start the server","error":"secret key is required"}
Error: 1 error occurred:
	* secret key is required

➜  bindplane-op git:(gen-secret-key-if-not-exist) ✗ docker run -it -e BINDPLANE_CONFIG_SECRET_KEY=$(uuidgen) observiq/bindplane-amd64 
{"level":"info","timestamp":"2023-03-09T23:35:43.140Z","message":"Using BBolt Storage","storageFilePath":"/data/storage"}
{"level":"info","timestamp":"2023-03-09T23:35:43.141Z","message":"disconnecting all agents on startup","count":0}
{"level":"info","timestamp":"2023-03-09T23:35:43.267Z","message":"Seeded ResourceTypes","dir":"destination-types","resou
```

I also tested against a fresh linux system with the `rpm` package. The default configuration (Before bindplane is started, and without running `init server`) looks like this with the comments removed:
```yaml
host: "127.0.0.1"
port: "3001"
serverURL: http://127.0.0.1:3001
logFilePath: /var/log/bindplane/bindplane.log
bindplaneHome: /var/lib/bindplane
server:
  remoteURL: ws://127.0.0.1:3001
  storageFilePath: /var/lib/bindplane/storage/bindplane.db
  downloadsFolderPath: /var/lib/bindplane/downloads
  #secretKey:
```

After start bindplane, it writes the config to look like this:
```yaml
server:
    storageFilePath: /var/lib/bindplane/storage/bindplane.db
    secretKey:  a33f73f7-ff22-4ea7-afbc-af2eaa07832a
    remoteURL: ws://127.0.0.1:3001
    env: production
    host: 127.0.0.1
    port: "3001"
    serverURL: http://127.0.0.1:3001
    username: admin
    password: admin
    logFilePath: /var/log/bindplane/bindplane.log
    syncAgentVersionsInterval: 1h0m0s
client:
    common:
        env: production
        host: 127.0.0.1
        port: "3001"
        serverURL: http://127.0.0.1:3001
        username: admin
        password: admin
        logFilePath: /var/log/bindplane/bindplane.log
command:
    output: table
```

Had I ran `init server`, I would have received the following config:
```yaml
host: 127.0.0.1
port: "3001"
serverURL: http://127.0.0.1:3001
username: admin
password: admin
logFilePath: /var/log/bindplane/bindplane.log
server:
    storageFilePath: /var/lib/bindplane/storage/bindplane.db
    secretKey: a33f73f7-ff22-4ea7-afbc-af2eaa07832a
    remoteURL: ws://127.0.0.1:3001
    sessionsSecret: da5ccf8d-6984-43ba-94ec-cd2204e48e50
```

The lack of consistency will probably be confusing. I know the way we read the config is probably the cause of that (the weird double unmarshalling). 

With that said, the written config works great.

BinaryFissionGames:
That's a great point. I updated this, it should only add the new secret key to your config (it'll strip out non-config keys, though). 

On that note, `downloadsFolderPath` and `bindplaneHome` aren't options you can specify in the config, apparently, but they are in our default config, confusingly enough.

jsirianni:
> That's a great point. I updated this, it should only add the new secret key to your config (it'll strip out keys that aren't valid, though).
> 
> On that note, `downloadsFolderPath` and `bindplaneHome` aren't options you can specify in the config, apparently, but they are in our default config, confusingly enough.

Looks better on my end now. I think `downloadsFolderPath` could go, that is a hold over from the original implementation of storing agent releases locally. I think `bindplaneHome` probably needs to stay but could be optional. IIRC, it is used as the default location to write files (like the log file) if not otherwise specified.

internal/cli/commands/serve/start.go
```diff
@@ -260,18 +262,44 @@ func (s *Server) createVersions(ctx context.Context, config *common.Server, st s
...
+	}
+
+	if configPath == "" {
+		// We can't rewrite the config with a new secret key if there is no config
+		return errors.New("secret key is required")
+	}
+
+	s.logger.Info("Generating secret key...")
+
+	c.SecretKey = uuid.NewString()
```
jsirianni, line 278: I know we use UUID in a few places, but I think technically we should be using a ULID.

BinaryFissionGames:
> I think bindplaneHome probably needs to stay but could be optional. IIRC, it is used as the default location to write files (like the log file) if not otherwise specified.

`bindplaneHome` is not [unmarshalled into on the config struct](https://github.com/observIQ/bindplane-op/blob/62d3524966db1e5ad844542df35683557c205f1d/common/config.go#LL112) because it's private on the config, hence why it's stripped away. This value is ignored in the config, as is. It can only be configured by the environment variable, I think (it's read here: https://github.com/observIQ/bindplane-op/blob/62d3524966db1e5ad844542df35683557c205f1d/internal/cli/commands/root.go#L44).

I think it kinda has to be that way, since it informs the default location to read the config from in the first place.

jsirianni (approved):
This is working great for me. We might want to create an additional ticket to address `bindplaneConfigHome` and the now removed downloads directory option.

PR #852, opened 2023-03-14
feat: Refactor CLI for extensibility

jmwilliams89:
### Proposed Change
- Refactors commands to take interfaces
- Renamed `cli.BindPlane` as `cli.Engine`
- Refactored `cli.BindPlane` to load interfaces
- Refactored `config.Config`
- Refactored config conversion logic
- Refactored`main.go`
- Added tests for each command

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


cli/commands/initialize/command.go
```diff
@@ -0,0 +1,79 @@
...
+	cmd := &cobra.Command{
+		Use:     "client",
+		Aliases: []string{"cli"},
+		Short:   "Initializes a new client installation",
+		Long:    ``,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if err := intializer.InitializeClient(cmd.Context()); err != nil {
+				return err
+			}
+			fmt.Println("\nInitialization complete!\nRun \"bindplane version\" to test the login.")
```
cpheps, line 73: Same nit here.

cli/commands/initialize/initializer.go
```diff
@@ -0,0 +1,284 @@
...
+	bytes, err := yaml.Marshal(i.cfg)
+	if err != nil {
+		return err
+	}
+
+	return ioutil.WriteFile(i.cfgPath, bytes, 0600)
+}
+
+// askQuestions asks the user questions and records their answers.
+func (i *initializer) askQuestions(questions []*Question, response any) error {
```
cpheps, line 116: Are there any restrictions on the response object?

cli/commands/initialize/initializer.go
```diff
@@ -0,0 +1,284 @@
...
+		stdioOpt := survey.WithStdio(i.stdio.In, i.stdio.Out, i.stdio.Err)
+		if err := survey.Ask(surveyQuestion, response, stdioOpt); err != nil {
+			return fmt.Errorf("%s input failed: %w", question.Name, err)
+		}
+	}
+	return nil
+}
+
+// Question is a question to ask the user during initialization.
+type Question struct {
```
cpheps, line 133: nit: I think these question and answer objects might be good to move into their own file. Just so it's obvious when looking that the questions live in `questions.go` or something like that.

cpheps (commented):
Made it through half of the commands. Submitting partial feedback and will continue reviewing.

cli/commands/profile/profiler.go
```diff
@@ -0,0 +1,292 @@
...
+// Profiler is an interface for managing BindPlane profiles
+type Profiler interface {
+	// GetProfileRaw returns the raw representation of a profile.
+	GetProfileRaw(ctx context.Context, name string) (string, error)
+	// GetCurrentProfileName returns the name of the currently used profile.
+	GetCurrentProfileName(ctx context.Context) (string, error)
+	// SetCurrentProfileName sets the name of the currently used profile.
+	SetCurrentProfileName(ctx context.Context, name string) error
+	// GetProfileNames returns a list of profile names.
+	GetProfileNames(ctx context.Context) ([]string, error)
```
cpheps, line 43: nit: Wondering if `ListProfileNames` is more descriptive than `GetProfileNames`. Could go either way just a thought.

cli/commands/profile/profiler.go
```diff
@@ -0,0 +1,292 @@
...
+// getCurrentProfileFile returns the path to the current profile file.
+func (p *profiler) getCurrentProfileFile() string {
+	return path.Join(p.folder, "current")
+}
+
+// profileExists returns true if the profile with the given name exists.
+func (p *profiler) profileExists(name string) bool {
+	filename := p.getProfilePath(name)
+	info, err := os.Stat(filename)
+	return err == nil && !info.IsDir()
```
cpheps, line 161: Do we need to check to make sure someone didn't maliciously create a `profile_name.yaml` dir?

cli/commands/validate/command.go
```diff
@@ -19,21 +19,40 @@ import (
...
+				if err := validator.ValidateClient(ctx); err != nil {
+					fmt.Fprint(writer, err)
+					os.Exit(1)
+				}
+			default:
+				fmt.Fprint(writer, "Must specify 'server' or 'client'\n")
+				os.Exit(1)
+			}
+
+			fmt.Fprint(writer, "configuration is valid\n")
```
cpheps, line 55: I don't think we need to `os.Exit(0)` after this I think that's the default functionality.

cli/commands/validate/command.go
```diff
@@ -19,21 +19,40 @@ import (
...
 // Command returns the BindPlane validate get cobra command
-func Command(bindplane *cli.BindPlane) *cobra.Command {
+func Command(validator Validator) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "validate",
 		Short: "validate the current profile",
 		Run: func(cmd *cobra.Command, args []string) {
-			if err := bindplane.Config.Validate(); err != nil {
-				fmt.Fprint(cmd.OutOrStdout(), err)
+			ctx := cmd.Context()
```
cpheps, line 31: We're using `RunE` everywhere else can we do that and return an error rather than doing `os.Exit`s?

cmd/bindplane/main.go
```diff
@@ -17,62 +17,96 @@
...
+		if err := os.MkdirAll(profilesFolder, 0750); err != nil {
+			return "", fmt.Errorf("failed to create profile folder: %w", err)
+		}
+		return profilesFolder, nil
+	}
+
+	// return an error if any other error occurs
+	if err != nil {
+		return "", fmt.Errorf("failed to access profile folder: %w", err)
+	}
```
cpheps, line 104: I think these two err checks can be condensed. I think `errors.Is` can take a nil error so we don't need to check.

Something like this maybe:

```go
switch {
  case errors.Is(err, os.ErrNotExits):
     // Do stuff
  case err != nil:
    // Do stuff
  case !info.IsDir():
    //  Do Stuff
}

File: config/client.go [hunk 9]
```diff
@@ -0,0 +1,75 @@
...
+
+// Client is the configuration for the client
+type Client struct {
+	// ServerURL is the URL that clients should use to contact the server.
+	ServerURL string `mapstructure:"serverURL" yaml:"serverURL,omitempty"`
+
+	// Username the basic auth username used for communication between client and server.
+	Username string `mapstructure:"username" yaml:"username,omitempty"`
+
+	// The basic auth password used for communication between client and server.
```
cpheps, line 30: Comment needs to start with the word `Password`

config/client.go
```diff
@@ -0,0 +1,75 @@
...
+	Password string `mapstructure:"password" yaml:"password,omitempty"`
+
+	// Output is the output type for the client
+	Output string `mapstructure:"output" yaml:"output"`
+
+	// TLSConfig is an optional TLS configuration for communication between client and server.
```
cpheps, line 36: Parameter name is just `TLS` not `TLSConfig`

config/server.go
```diff
@@ -0,0 +1,185 @@
...
+	// Host is the Host to which the server will bind.
+	Host string `mapstructure:"host" yaml:"host,omitempty"`
+
+	// Port is the Port on which the server will serve.
+	Port string `mapstructure:"port" yaml:"port,omitempty"`
+
+	// Username the basic auth username used for communication between client and server.
+	Username string `mapstructure:"username" yaml:"username,omitempty"`
+
+	// The basic auth password used for communication between client and server.
```
cpheps, line 58: Comment needs to start with the word `Password`

config/server.go
```diff
@@ -0,0 +1,185 @@
...
+	Password string `mapstructure:"password" yaml:"password,omitempty"`
+
+	// TLSConfig is an optional TLS configuration for communication between client and server.
```
cpheps, line 61: `TLS` instead of `TLSConfig`

cli/commands/apply/applier.go
Mrod1598, line 1: Missing protocol? 
image

cli/commands/initialize/command.go
Mrod1598, line 1: We were generating the secret key if none was given, do we actually want this to be required? 
image

cli/commands/validate/command.go
Mrod1598, line 1: command doesn't seem to working 
image
profile: 
|
image

Mrod1598 (changes requested):

cli/commands/apply/applier.go
```diff
@@ -0,0 +1,69 @@
...
+
+// NewApplier returns a new Applier.
+func NewApplier(client client.BindPlane) Applier {
+	return &applier{
+		client: client,
+	}
+}
+
+// applier is the default implementation of Applier.
+type applier struct {
```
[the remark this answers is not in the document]
jmwilliams89: Good call. I'll change all the default implementations of all these interfaces to be prefaced with default. I was just going for simplistic names at the time, since they're not exported. But I think this makes it more obvious for anyone reading the code.

cli/commands/apply/applier_test.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/observiq/bindplane-op/model"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+)
+
+func TestApplyResourcesFromFiles(t *testing.T) {
+	testCases := []struct {
+		name             string
+		clientFunc       func() client.BindPlane
+		setupFunc        func(path string)
```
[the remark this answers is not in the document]
jmwilliams89: Good change. There are probably multiple tests that require this pattern.

cli/commands/delete/deleter.go
```diff
@@ -0,0 +1,89 @@
...
+	"fmt"
+
+	"github.com/observiq/bindplane-op/client"
+	"github.com/observiq/bindplane-op/model"
+)
+
+// Deleter is an interface for deleting resources.
+type Deleter interface {
+	// DeleteResource deletes a single resource.
+	DeleteResource(ctx context.Context, kind model.Kind, id string) error
```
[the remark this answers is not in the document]
jmwilliams89: Based on the previous design, it looks like the two functions are different. `DeleteResource` supports all resource types, whereas `DeleteResources` supports just agents. I think this was because the client doesn't have any functions for deleting multiples of other resources. We could support this, if we wanted to introduce a loop here that calls individual deletes for the other resource types.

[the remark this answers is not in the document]
jmwilliams89: I handled this by adding the aforementioned loop.

cli/commands/get/getter.go
```diff
@@ -0,0 +1,247 @@
...
+// Getter is an interface for getting and printing resources.
+type Getter interface {
+	// GetResource gets and prints a single resource.
+	GetResource(ctx context.Context, kind model.Kind, id string) error
+	// GetSomeResources gets and prints a list of resources.
+	GetSomeResources(ctx context.Context, kind model.Kind, ids []string) error
+	// GetAllResources gets and prints all resources of a given type.
+	GetAllResources(ctx context.Context, kind model.Kind, queryOpts ...client.QueryOption) error
+	// GetEveryResource gets and prints all resources of all types.
+	GetEveryResource(ctx context.Context) error
```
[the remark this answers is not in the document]
jmwilliams89: I went with `GetResource`, `GetResources`, `GetResourcesOfKind`, and `GetAllResources`.

cli/commands/initialize/initializer.go
jmwilliams89: No. The underlying survey library accepts `response interface{}` when asking a question. It uses tags to populate any structs given to it.

cli/commands/initialize/initializer.go
jmwilliams89: Good idea. I think it's best to make a `questions` package under the `initialize` package then. This would imply to any outside consumers that the questions are intended for use in the initialize command.

cli/commands/profile/profiler.go
jmwilliams89: I'm going to keep `GetProfileNames`, since we've been using this pattern elsewhere (see `getter.go`).

cli/commands/profile/profiler.go
jmwilliams89: This was the preexisting behavior. I agree. It seems unrealistic. I'll pull it out.

cli/commands/validate/command.go
jmwilliams89: I actually don't think any of the `os.Exit`'s in this file are necessary. This was the preexisting behavior of the command before, so I kept it. But now that I'm looking closer, I noticed that `cobra.CheckErr` will handle error code 1 all on its own. I'm going to remove all the manual exits and change the command to return errors like the other command.
image

cli/commands/validate/command.go
jmwilliams89: Great minds think alike. See above. I just noticed this as well.

cmd/bindplane/main.go
jmwilliams89: That's funny. I had this at one point, but I backed out, because I thought `errors.Is` might act strangely with nil. I'll switch it back to this behavior.

cli/commands/initialize/initializer.go
cpheps: I like this changes it breaks things up nicely.

cli/engine.go
```diff
@@ -0,0 +1,547 @@
...
+		}
+		name = currentProfileName
+	}
+
+	path := path.Join(b.profilesFolder, fmt.Sprintf("%s.yaml", name))
+	return b.LoadConfig(context.Background(), path)
+}
+
+// GetProfileRaw returns the raw profile data.
+func (b *Engine) GetProfileRaw(ctx context.Context, name string) (string, error) {
```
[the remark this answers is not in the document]
jmwilliams89: The previous implementation was reading from file, unmarshaling to the profile struct,  then marshaling back to bytes and printing that to console. That's why any fields that didn't exist on the struct were stripped away before printing. This felt like an unnecessary optimization, when we can simply read from the file. I think the question is, do we want to enforce that the profile is valid and only shows relevant fields when the user calls `profile get`. I can go either way on this.

cli/commands/profile/command.go
```diff
@@ -0,0 +1,210 @@
...
+// Command returns the BindPlane profile cobra command.
+func Command(profiler Profiler) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "profile",
+		Short: "Profile commands.",
+		Long:  "Profile commands for managing BindPlane application configuration",
+	}
+
+	cmd.AddCommand(
+		GetCommand(profiler),
```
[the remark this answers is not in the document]
jmwilliams89: Following up on this. I'm going to remove `bindplanehome` from the configs. This was a holdover from the previous design. Since this value is always either the env variable or the default path, we can simply use a common func for this.

cli/commands/apply/applier.go
jmwilliams89: This ended up being an issue with an ambiguous selector when viper read in the config. Rather than embedding `Client` and `Server`, we can avoid this by directly naming those fields.

cli/commands/validate/command.go
jmwilliams89: I went ahead and removed the UUID requirement. Secret key is no longer validated in this manner.

Mrod1598 (approved):
Working and Code looks good.

cpheps (approved):

PR #853, opened 2023-03-14
feat(sourcetype): Add cloudflare source

dehaansa:
### Proposed Change
Add support for the Cloudflare LogPush receiver. 

~~Dependent on https://github.com/observIQ/observiq-otel-collector/pull/976~~ and release of collector version 1.21.0 of the collector.

image

image

image

##### Checklist
- [X] Changes are tested
- [x] CI has passed


resources/source-types/cloudflare.yaml
```diff
@@ -0,0 +1,73 @@
...
+  icon: /icons/sources/cloudflare.svg
+  description: Receive logs from Cloudflare LogPush jobs
+spec:
+  version: 0.0.1
+  supported_platforms:
+    - macos
+    - linux
+    - windows
+  parameters:
+    - name: listen_endpoint
```
jsirianni, line 15: Can we split these up into a string and int params? Similar to https://github.com/observIQ/bindplane-op/blob/release/v1.11.0/resources/source-types/otlp.yaml#L15

resources/source-types/cloudflare.yaml
```diff
@@ -0,0 +1,73 @@
...
+      type: string
+      required: true
+      default: ""
+
+    - name: listen_tls_cert_file
+      label: Cert File Location
+      description: Local path to the TLS cert file.
+      type: string
+      required: true
+      default: ""
```
jsirianni, line 33: Is TLS required? (I imagine it is).

I think we should be consistent with TLS parameter names. The latest iteration of the OTLP source looks like this https://github.com/observIQ/bindplane-op/blob/fix/otlp-resource-tls/resources/source-types/otlp.yaml#L44

dehaansa: happy to make it more consistent. it is required, yes.

dehaansa:
image

Updates based on feedback from @jsirianni

jsirianni (approved):

resources/source-types/cloudflare.yaml
```diff
@@ -0,0 +1,77 @@
...
+    - name: secret
+      label: Shared Secret
+      description: "String used to validate that message is coming from an expected source. If set, it must be also be configured in the LogPush job's destination field like the following: `\"destination_conf\": \"https://example.com?header_X-CF-Secret=abcd1234\"`"
+      type: string
+      default: ""
+      required: false
+
+    - name: attributes
+      label: Attributes
+      description: Any Cloudflare LogPush field names provided as keys will be mapped to metric attributes using the map value as the attribute name.
```
jsirianni, line 48:
```suggestion
      description: Any Cloudflare LogPush field names provided as keys will be mapped to log record attributes using the map value as the attribute name.
```

dehaansa:
updated based on @antonblock feedback.

image

antonblock (approved):

jsirianni (approved):

PR #871, opened 2023-03-20
feat: Convert from legacy config

jmwilliams89:
### Proposed Change
- Moved the old config structs to a legacy package
- Removed unused legacy structs
- Added deprecation comments for all legacy structs and types
- Added a function for converting a legacy config file
- Added legacy conversion logic to `LoadConfig`

##### Checklist
- [x] Changes are tested
- [x] CI has passed


common/common_test.go
```diff
@@ -0,0 +1,48 @@
...
+	require.Equal(t, expectedPath, home)
+}
+
+func TestGetHomeWithEnv(t *testing.T) {
+	value := os.Getenv(HomeEnv)
+	if value != "" {
+		defer os.Setenv(HomeEnv, value)
+	}
+
+	err := os.Setenv(HomeEnv, "test-home")
```
cpheps, line 36: I'm not as sure what we're doing here but are we actually using a directory called `test-home` or is it just for testing reading the env?

If we're using a real directory maybe we should use `t.TmpDir`. If not this is fine.

jmwilliams89: This is just a test that validates if we set the home environment variable, it will be used for the home path, instead of the default.

PR #872, opened 2023-03-21
feat: Standardized tracer package, build, and config

jmwilliams89:
### Proposed Change
- Pulled the tracing config into the config package.
  - This will help standardize where all config structs reside.
- Updated the tracer implementation and build process to mirror how `Store` is built. 
- Renamed `OpenTelemetryTracing` to `OTLPTracing`
  - `OpenTelemetryTracing` was misleading, since both tracer implementations are actually using the otel library. The difference is primarily in how the spans are exported. This new name should reflect that distinction.
- Renamed the `trace` package to `tracer`, since this interface is the primary purpose of the package.
- Simplified names in the `tracer` package to remove stuttering.

##### Checklist
- [x] Changes are tested
- [x] CI has passed


tracer/otlp.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	_, _, err := net.SplitHostPort(o.cfg.Endpoint)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse gRPC endpoint: %v", err)
+	}
+
+	conn, err := grpc.DialContext(ctx, o.cfg.Endpoint, dialOpts...)
+	if err != nil {
+		return nil, fmt.Errorf("failed to dial otlp endpoint: %v", err)
```
antonblock, line 94:
```suggestion
		return nil, fmt.Errorf("failed to dial otlp endpoint: %w", err)
```

cpheps (approved):

PR #874, opened 2023-03-21
feat(processortype): Batch processor

jsirianni:
### Proposed Change

Added a processor type for the [batch](https://github.com/open-telemetry/opentelemetry-collector/tree/main/processor/batchprocessor) processor. It can be useful to configure batching outside of what is included in some of the destination types.

Screenshot from 2023-03-21 13-36-56

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


resources/processor-types/batch.yaml
```diff
@@ -0,0 +1,36 @@
...
+    - name: timeout
+      label: Timeout
+      type: string
+      default: 200ms
+      description: "Time duration after which a batch will be sent regardless of size. Example: 2s (two seconds)"
+      required: true
+
+  logs+metrics+traces:
+    processors: |
+      - batch/2:
```
cpheps, line 33: If I add two of these won't they conflict since the `/2` is part of the unique name?

jsirianni: the /2 should be fine but it was a mistake. It should just be `batch`. Fixed.

BinaryFissionGames (approved):

PR #885, opened 2023-03-22
feat(UI): Password option for string parameters.

dsvanlani:
### Proposed Change
- Provides new ParameterOption, `password: true`.
- Allows string parameter inputs to hide their text if password option is set.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


Mrod1598 (approved):

jsirianni (approved):
Working great. NIT: I do wonder if we should call it "sensitive" instead of "password", considering it can be used for things like New Relic License Key.

PR #886, opened 2023-03-23
feat: Refactored CLI into a builder pattern

jmwilliams89:
### Proposed Change
- Added a builder interface for each command. 
  - This helps reduce the amount of overhead when loading the CLI for a command.
  - Components will only be built when required for a command.
  - This should make the struct stateless and remove any concerns about order of operations when loading.
- Moved config creation into the new factory.
  - Previously, the config was created in the main func and passed into the CLI.
  - The new design creates a new config struct each time the config is loaded.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


cli/factory.go
```diff
@@ -0,0 +1,346 @@
...
+
+	tr, err := f.BuildTracer(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load tracer: %w", err)
+	}
+
+	return serve.NewServer(&f.cfg.Server, st, tr, logger), nil
+}
+
+// BuildLogger builds a logger.
```
Mrod1598, line 252: looked to me like we were building a new logger for everything, same with client, etc. think it might be worth at least updating the comment. because some of the builders do create a new version of `X` and some don't. like `BuildApplier` creates a new applier every time
```suggestion
// BuildLogger builds a logger, if it doesn't already exist.
```

Mrod1598 (approved):

PR #892, opened 2023-03-28
refactor:  Query options and store Updates

Mrod1598:
### Proposed Change
- Expose Query option fields and methods
- Use Interface for managing events on the eventbus for updates
- remove pointer generics in eventbus.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


eventbus/source.go
```diff
@@ -204,6 +204,8 @@ func NewSource[T any]() Source[T] {
 	}
 }
 
+var _ Source[int] = (*source[int])(nil)
```
cpheps, line 207: Is this required?

cpheps (approved):

eventbus/source.go
Mrod1598: no not necessarily just was using it to verify it did implement it because a test was failing saying it wasn't. I'll remove it

antonblock (approved):

PR #893, opened 2023-03-29
feat(source-types): Add parse_to param to plugins supporting it

antonblock:
### Proposed Change

- Add `parse_to` field as an advanced parameter to source types that support it
- Default to `body`, which the collector does

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


resources/source-types/syslog.yaml
```diff
@@ -215,6 +225,7 @@ spec:
           {{ if and (.enable_non_transparent_framing_trailer) (eq .connection_type "tcp") (eq .protocol "rfc5424")}}
           non_transparent_framing_trailer: {{ .non_transparent_framing_trailer }}
           {{ end}}
+          parse_to: {{ .parse_to }}
```
BinaryFissionGames, line 228: This is the actual syslog receiver, not the collector plugin, which doesn't seem to have this option: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/syslogreceiver#configuration

Probably needs to be implemented the same way we implement it in the plugin:
https://github.com/observIQ/observiq-otel-collector/blob/f79a4eefd43b244560a0494a1e63c6973edbf883/plugins/syslog_logs.yaml#L116

BinaryFissionGames (changes requested):

BinaryFissionGames (approved):

PR #896, opened 2023-03-30
feature(filelog): Parse timestamps and severity

antonblock:
### Proposed Change

- Update `file_log` with fields to parse timestamp and severity from log entries
- Timestamp initially supporting `ISO8601`, `RFC3339`, `Epoch`, or manual `strptime`
- Severity is mapped using default mapping

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


resources/source-types/file_log.yaml
```diff
@@ -88,6 +88,93 @@ spec:
...
+          operator: equals
+          value: true
+        - name: parse_timestamp_format
+          operator: equals
+          value: Manual
+      documentation:
+        - text: Supported Layout Directives
+          url: https://github.com/observiq/ctimefmt/blob/3e07deba22cf7a753f197ef33892023052f26614/ctimefmt.go#L63
+
+    - name: parse_severity
```
jsirianni, line 159: I think we should require `parse_format != none` in order to parse a sev field.

resources/source-types/file_log.yaml
```diff
@@ -88,6 +88,93 @@ spec:
           operator: equals
           value: specify line end
 
+    - name: parse_timestamp
```
jsirianni, line 91: I think we should require parse_format != none in order to parse a timestamp field.

resources/source-types/file_log.yaml
```diff
@@ -210,11 +297,41 @@ spec:
...
+                {{ if eq .parse_timestamp_format "RFC3339" }}layout: '%Y-%m-%dT%H:%M:%S.%f'{{ end }}
+                {{ if eq .parse_timestamp_format "Manual" }}layout: '{{ .manual_timestamp_format }}'{{ end }}
+                {{ if eq .parse_timestamp_format "Epoch" }}
+                layout_type: epoch
+                layout: {{ .epoch_timestamp_format }}
+                {{ end }}
+            {{ end }}
+            {{ if .parse_severity }}
+              severity:
+                parse_from: {{ .parse_to }}.{{ .severity_field }}
```
jsirianni, line 333: It may be out of scope or not currently possible, but I think we should support severity mappings. For example, parsing severity from an http status code would mean something like this:
```yaml
          severity:
            parse_from: "attributes.status"
            mapping:
              info: 2xx
              info2: 3xx
              warn: 4xx
              error: 5xx
```

resources/source-types/file_log.yaml
```diff
@@ -88,6 +88,93 @@ spec:
...
+        - Manual
+      relevantIf:
+        - name: parse_timestamp
+          operator: equals
+          value: true
+
+    - name: epoch_timestamp_format
+      label: Epoch Layout
+      description: The layout of the epoch-based timestamp.
+      type: string
```
jsirianni, line 128: I think we can use an enum here if we know all of the possible values

jsirianni (commented):
Seems to be working great, left a few comments that I think we could resolve before more final testing.

jsirianni (approved):

Mrod1598 (approved):

PR #909, opened 2023-04-04
feat: Use overrides when updating profiles

jmwilliams89:
### Proposed Change
- Updated the profile command to use overrides when setting values
- Updated config overrides to be created with functions for more consistency
- Moved logging config into the `config` package

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


config/auth.go
```diff
@@ -0,0 +1,73 @@
...
+		return errors.New("password must be set")
+	}
+
+	if c.SecretKey == "" {
+		return errors.New("secret key must be set")
+	}
+
+	if c.SessionSecret == "" {
+		return errors.New("session secret must be set")
+	}
```
cpheps, line 70: Is this required? I don't think I have it in any of my local configs and everything works. Though I'm not sure what it's for.

cpheps (approved):

Mrod1598 (approved):

Mrod1598:
Actually this flag seems to be missing
image

PR #927, opened 2023-04-10
Fix: Snapshot Summary metrics

dsvanlani:
### Proposed Change
We were getting an error marshaling the response for the `snapshot` query into JSON.  Previously the value was a `map[float65]any`.  So now we just make this a `map[string]string` and then `JSON.stringify` the value on the react side.

Changes:
- Return value of Summary metric as `map[string]any`.
- Stringify value display for the `MetricsRecordRow`.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


antonblock (approved):

jsirianni (approved):
Working great in our env, which was experiencing this issue.

PR #945, opened 2023-04-14
fix: Inject commit information at build time

BinaryFissionGames:
### Proposed Change
* Inject the commit into the version struct when building (this field was present, but never populated)
* Print both the tag and commit from `bindplanectl version`

This information is pretty useful when trying to determine the exact build, since the tag isn't necessarily pinned to a commit, and some methods of building locally may not be accurate with the tag.

A test build with the version info embedded can be created with `make build`.

```
$ ./dist/bindplanectl_darwin_amd64_v1/bindplanectl version
client: v1.12.2 (4d89ea969247999548def82402cc0889d465babc)
server: v1.12.2 (4d89ea969247999548def82402cc0889d465babc)
```

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


jsirianni (approved):

jsirianni:
When building w/ Goreleaser, I see this
```
client: v1.12.2-SNAPSHOT-7dc9e75 (7dc9e7594ff144edd06f34f0ca501c04da3f7c09)
```
Somewhat redundant but I do like it better. Im only mentioning it because it seems to differ with your results. Im not sure why.

antonblock (approved):

jsirianni (approved):

PR #964, opened 2023-04-19
feat(agents): Add k8s daemonset deployment option

antonblock:
### Proposed Change

- Add `kubernetes` platform for configurations, agents

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


internal/rest/install_command.go
```diff
@@ -152,21 +160,50 @@ func (p *installCommandParameters) installerURL() string {
...
+func (p *installCommandParameters) installCommand() (string, error) {
 	switch p.platform {
 	case windowsAmd64:
 		return fmt.Sprintf(`msiexec /i "%s" /quiet%s`,
 			p.installerURL(),
 			p.args(),
-		)
+		), nil
+	case kubernetesDaemonset:
+		t, err := template.New("deployment").Parse(k8sDaemonsetChart)
```
cpheps, line 171: Is it worth parsing this once or is it fine doing it on demand? We're already storing the entire template i memory so I don't think it's a big deal either way. More just wanted your thoughts.

internal/rest/helm_charts.go
```diff
@@ -0,0 +1,170 @@
+// Copyright observIQ, Inc.
```
cpheps, line 1: Should this k8s stuff go in a sub package under rest? The rest package is pretty large I'm not sure if it's useful to separate out the k8s logic.

jsirianni, line 1: I think we can rename helm_chart.go to something like `kubernetes.go`. We do not have a helm install option right now, just a regular kubernetes manifest shown here. We may, in the future, have a second option for installing agents, using the opentelemetry community's helm chart.

cpheps: Do I need helm to install this or is that what you're saying @jsirianni that it's just a kube config?

jsirianni:
I am having an issue with the install agent page.

Steps:
1. Apply source types which have the supported platform `kubernetes-daemonset`
2. Build a config for kubernetes w/ the kubernetes source types

Even after a refresh, I am unable to select the config on the agent install page.

internal/rest/helm_charts.go
cpheps: I think it might be worth adding instructions. We're just generating a k8s config file and it may not be obvious to everyone what to do with that.

cpheps:
It seems to be working as expected.

internal/rest/helm_charts.go
antonblock: I think it still makes sense to be in `rest`

internal/rest/install_command.go
antonblock: Yeah I think if this was expensive I'd want a way to only do it once, but for now I think I prefer keeping it all contained in the `installCommand` method

cpheps (approved):

jsirianni (approved):

jsirianni (approved):

PR #966, opened 2023-04-20
Set proper kubernetes platform in k8s source types

jsirianni:
### Proposed Change

This PR is for the epic/k8s-resources feature branch, which contains four source types.
- Setting `kubernetes-daemonset` in the node and container source types, pr this pr https://github.com/observIQ/bindplane-op/pull/964
- Setting `kubernetes-deployment` with the assumption that we will use that as a platform, can always change it later.

##### Checklist
- [x] Changes are tested
- [x] CI has passed


cpheps (approved):
Do we need to bump the versions on any of these?

antonblock (approved):

jsirianni:
> Do we need to bump the versions on any of these?

I think we are okay to skip bumping the version until k8s resources are released.

PR #978, opened 2023-04-24
fix(config): Don't show Add button for container configs

antonblock:
### Proposed Change

Don't allow adding agents from the config page if the config is for containers. We intend that to be done from the install page

##### Checklist
- [x] Changes are tested
- [x] CI has passed


dsvanlani (approved):

jsirianni (approved):
Looks good. I think in the future, we could re-add the add option and have it link to the agent install page.

PR #987, opened 2023-04-25
feat(sourcetype): Splunk HEC source type

jsirianni:
### Proposed Change

Added a new source type which uses the [Splunk HEC receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/splunkhecreceiver)

This is a minimal source type, with support for the following receiver params
- endpoint
- access_token_passthrough
- server side tls

Tested using the HEC destination type.

Screenshot from 2023-04-25 16-43-47

##### Checklist
- [x] Changes are tested
- [x] CI has passed


resources/source-types/splunk_hec.yaml
```diff
@@ -0,0 +1,89 @@
+apiVersion: bindplane.observiq.com/v1
+kind: SourceType
+metadata:
+  name: splunkhec
+  displayName: Splunk HTTP Event Collector (HEC)
+  icon: /icons/destinations/splunk.svg
+  description: Receive logs from a Splunk HTTP Event forwarder.
+spec:
+  parameters:
```
BinaryFissionGames, line 9: Should we have the `version` and `supportedPlatforms` here?

resources/source-types/splunk_hec.yaml
```diff
@@ -0,0 +1,89 @@
...
+        - name: enable_tls
+          operator: equals
+          value: true
+      advancedConfig: true
+
+  logs:
+    receivers: |
+      - splunk_hec:
+          endpoint: "{{ .listen_ip }}:{{ .listen_port }}"
+          access_token_passthrough: "{{ .access_token_passthrough }}"
```
BinaryFissionGames, line 71: nit: I think we should unquote this field, since it's a boolean value and not a string. I'm a little surprised it works this way, but I think the collector has special unmarshalling logic to unmarshal string to bool, so it works as expected either way.

resources/source-types/splunk_hec.yaml
```diff
@@ -0,0 +1,94 @@
...
+    - name: listen_port
+      label: Listening Port
+      description: The port to bind to and receive HTTP events. Collector must be running as root (Linux) or Administrator (windows) when binding to a port below 1024.
+      type: int
+      default: 8088
+      required: true
+
+    - name: access_token_passthrough
+      label: Access Token Passthrough
+      description: Whether to preserve incoming access token (Splunk header value) as "com.splunk.hec.access_token" metric resource label.
```
shazlehu, line 31: Maybe we don't support this, but this description doesn't appear in the UI. Not in the scope of this PR, but the link for the token field in the Splunk destination redirects to the top level docs, https://docs.splunk.com/Documentation

shazlehu (approved):

BinaryFissionGames (approved):

PR #994, opened 2023-04-27
fix(plugin): Resolve Coralogix pipeline toggles

jsirianni:
### Proposed Change

broke up `logs+metrics+traces` into individual pipelines to resolve an issue where toggling a pipeline off would result in that pipeline being deployed anyone, but without the endpoint for that pipeline.

Example: Turning logs off would result in a log pipeline using the exporter, but the log endpoint param is not set and we see this error:
```
Failed to apply config changes: failed to reload config: collector.yaml: collector failed to restart: failed to build pipelines: failed to create "coralogix/coralogix" exporter for data type "logs": coralogix exporter config requires `logs.endpoint` configuration
```

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


resources/destination-types/coralogix.yaml
```diff
@@ -232,13 +232,77 @@ spec:
...
     extensions: |
+      {{ if .enable_logs }}
       {{ if and .persistent_queue_enabled .sending_queue_enabled }}
       - file_storage:
           directory: {{ .persistent_queue_directory }}
       {{ end }}
+      {{ end }}
+    exporters: |
+      {{ if .enable_logs }}
+      - coralogix:
```
BinaryFissionGames, line 245:
```suggestion
      - coralogix/logs:
```

resources/destination-types/coralogix.yaml
```diff
@@ -232,13 +232,77 @@ spec:
...
+    extensions: |
+      {{ if .enable_metrics }}
+      {{ if and .persistent_queue_enabled .sending_queue_enabled }}
+      - file_storage:
+          directory: {{ .persistent_queue_directory }}
+      {{ end }}
+      {{ end }}
     exporters: |
+      {{ if .enable_metrics }}
       - coralogix:
```
BinaryFissionGames, line 306:
```suggestion
      - coralogix/metrics:
```

BinaryFissionGames (changes requested):
My agent errrors when I have two switches enabled, I think it's because the exporters need to have different names, otherwise only one of the exporters takes?

jsirianni:
> 

Yeah you are right. It worked for me somehow, but I am now noticing that the trace pipeline is omitted at all times. Ill sort it out.

BinaryFissionGames (approved):

antonblock (approved):

PR #1000, opened 2023-04-28
feat(CockroachDB)

BominRahmani:
A source for the cockroachDB plugin here (https://github.com/observIQ/observiq-otel-collector/pull/1039)
##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


jsirianni:
Does this source type depend on a collector plugin pull request? If so, can you link it here? Thanks!

resources/source-types/cockroach.yaml
```diff
@@ -0,0 +1,311 @@
...
+    relevantIf:
+      - name: enable_metrics
+        operator: equals
+        value: true
+
+  - name: port
+    label: Port
+    description: The TCP port of the CockroachDB system.
+    type: int
+    default: 8080
```
jsirianni, line 37: Is there documentation on this port? If so, can we link it using [this pattern](https://github.com/observIQ/bindplane-op/blob/feature/cockroachdb-source/resources/destination-types/grafana_cloud.yaml#L14)?

jsirianni: Similarly, lets use the external documentation "read more" on any parameter which has associated documentation, like username / password or information on each log type. 

This is a newer feature, so you may not have seen it in previous source types.

resources/source-types/cockroach.yaml
```diff
@@ -0,0 +1,311 @@
...
+    advancedConfig: true
+
+  # Logging
+  #
+  - name: enable_logs
+    label: Enable_Logs
+    description: Enable to collect logs.
+    type: bool
+    default: true
+  - name: enable_health_log
```
jsirianni, line 83: nit, lets be consistent with spacing between each param.
```suggestion
    default: true
  
  - name: enable_health_log
```

resources/source-types/cockroach.yaml
```diff
@@ -0,0 +1,311 @@
...
+
+            enable_health_log: {{ .enable_health_log }}
+            health_log_path:
+              {{ range $fp := .health_log_path }}
+              - '{{ $fp }}'
+              {{ end }}
+      {{ end }}
+
+    processors: |
+      - resourcedetection:
```
jsirianni, line 307: We should use this processor in the metrics pipeline as well.

jsirianni (commented):
Looks great, have not tried it out yet but will when the PR is finalized. Left a few comments.

BominRahmani:
> Does this source type depend on a collector plugin pull request? If so, can you link it here? Thanks!

Hey Joe! Thanks for the comments. I took care of most of them.
Here is the link to the PR this is dependent on. https://github.com/observIQ/observiq-otel-collector/pull/1039
I will also update the description of this PR to include this link!

resources/source-types/cockroach.yaml
```diff
@@ -0,0 +1,518 @@
...
+    default: ""
+    documentation:
+      - text: Read more
+        url: https://www.cockroachlabs.com/docs/v23.1/cockroach-cert#client-key-and-certificates
+    options:
+        gridColumns: 12
+    relevantIf:
+      - name: enable_metrics
+        operator: equals
+        value: true
```
jsirianni, line 123: Lets follow [this](https://github.com/observIQ/bindplane-op/blob/main/resources/source-types/nginx.yaml#L52) pattern for TLS parameters.
- TLS options should be gated behind an "enable tls" toggle
- Cert file and key file options should make it clear that they are for mutual tls

resources/source-types/cockroach.yaml
```diff
@@ -0,0 +1,518 @@
...
+      - text: Read more
+        url: https://www.cockroachlabs.com/docs/v23.1/cockroach-cert#client-key-and-certificates
+    options:
+        gridColumns: 12
+    relevantIf:
+      - name: enable_metrics
+        operator: equals
+        value: true 
+
+  - name: server_name
```
jsirianni, line 125: The read more link indicates that this is a TLS option. It should be nested under the "enable tls" toggle mentioned in my previous comment. We can probably remove the "advanced: true" option here.

resources/source-types/cockroach.yaml
```diff
@@ -0,0 +1,518 @@
...
+    description: How often (seconds) to scrape for metrics.
+    type: int
+    default: 60
+    relevantIf:
+      - name: enable_metrics
+        operator: equals
+        value: true
+    advancedConfig: true
+  
+  - name: insecure_skip_verify
```
jsirianni, line 151: This should be nested under the "enable tls" toggle mentioned in my previous comment. We can probably remove the "advanced: true" option here.

jsirianni (changes requested):

resources/source-types/cockroach.yaml
BominRahmani: I kept the advanedConfig: true to keep it bundled with the TLS option, since the tls option seems to fall under an advancedConfig: true in the pattern you linked as well as in the elastic search.

resources/source-types/cockroach.yaml
BominRahmani: I kept the advanedConfig: true to keep it bundled with the TLS option, since the tls option seems to fall under an advancedConfig: true in the pattern you linked as well as in the elastic search.

resources/source-types/cockroach.yaml
```diff
@@ -0,0 +1,534 @@
...
+    documentation:
+      - text: Read more
+        url: https://datatracker.ietf.org/doc/html/rfc4366#section-3.1
+    relevantIf:
+      - name: enable_tls
+        operator: equals
+        value: true
+    advancedConfig: true
+
+  - name: collection_interval
```
jsirianni, line 151: Lets re-arrange the params so all TLS options are grouped. Right now, collection interval sits between server name and skip verification.

Screenshot from 2023-06-15 13-59-52

jsirianni (commented):
I think this looks good, left one more comment.

resources/source-types/cockroach.yaml
BominRahmani: image

resources/source-types/cockroach.yaml
```diff
@@ -0,0 +1,534 @@
...
+      sectionHeader: true
+    relevantIf:
+      - name: enable_metrics
+        operator: equals
+        value: true
+    advancedConfig: true
+
+  - name: ca_file_path
+    label: CA File Path
+    description: File path for CA certification file for CockroachDB (only needed if you have a secure cluster)
```
BinaryFissionGames, line 93:
```suggestion
    description: File path for CA certification file for CockroachDB (only needed if you have a secure cluster).
```

resources/source-types/cockroach.yaml
```diff
@@ -0,0 +1,534 @@
...
+        url: https://www.cockroachlabs.com/docs/v23.1/cockroach-cert#client-key-and-certificates
+    relevantIf:
+      - name: enable_tls
+        operator: equals
+        value: true 
+    advancedConfig: true
+
+  - name: server_name
+    label: Server Name
+    description: Name of server
```
BinaryFissionGames, line 138:
```suggestion
    description: The name of the server.
```

resources/source-types/cockroach.yaml
```diff
@@ -0,0 +1,534 @@
...
+  #
+  - name: enable_logs
+    label: Enable_Logs
+    description: Enable to collect logs.
+    type: bool
+    default: true
+  
+  - name: enable_health_log
+    label: Enable Health Logs
+    description: Enable to collect health logs
```
BinaryFissionGames, line 186: Could we add a period to the end of all these descriptions, just so they are consistent?

resources/source-types/cockroach.yaml
```diff
@@ -0,0 +1,534 @@
...
+          path: $OIQ_OTEL_COLLECTOR_HOME/plugins/cockroachdb_metrics.yaml
+          parameters:
+            endpoint: {{ .hostname }}:{{ .port }}
+            username: {{ .username | quote}}
+            password: {{ .password | quote}}
+            ca_file_path: {{ .ca_file_path | quote}}
+            cert_file_path: {{ .cert_file_path | quote}}
+            key_file_path: {{ .key_file_path | quote}}
+            insecure_skip_verify: {{ .insecure_skip_verify }}
+            server_name: {{ .server_name | quote}}
```
BinaryFissionGames, line 464: We should only fill these parameters in if `enable_tls` is true, otherwise if you fill in the tls parameters and uncheck the tls enable, tls still applies.

resources/source-types/cockroach.yaml
```diff
@@ -0,0 +1,542 @@
...
+      - text: Read more
+        url: https://www.cockroachlabs.com/docs/v22.2/create-user#create-a-user
+    relevantIf:
+      - name: enable_metrics
+        operator: equals
+        value: true
+
+  - name: password
+    label: Password
+    description: The password to use when connecting to CockroachDB. TLS must be configured in advance if this field is set.
```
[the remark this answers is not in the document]
BominRahmani: Glad to hear it!

BinaryFissionGames (approved):

PR #1003, opened 2023-05-01
feat: Require supportedPlatforms field and filter list in UI

dsvanlani:
### Proposed Change
- Validates that the `spec.supportedPlatforms` field is specified for Source Types.
- Filter list of `ResourceTypes` by a new `platform` prop in the `SelectView` used in the `NewResourceDialog` component.

We are not adding `k8s` or `OpenShift` support to any existing Source Type, and Custom source type will not be part of k8s for this pass.

image

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


model/resource_type.go
```diff
@@ -369,6 +371,67 @@ func (s *ResourceTypeSpec) validate(kind Kind, errs validation.Errors) {
 	s.Traces.validateTemplates(errs, "traces", params)
 }
 
+const (
+	platformWindows             = "windows"
```
jsirianni, line 375: I wonder if we should refactor this and `internal/rest/install_command.go`'s constants, so platforms are defined in one place. Probably not now, but perhaps a ticket for the future.

```go
type supportedPlatform string

const (
	kubernetesDaemonset  supportedPlatform = "kubernetes-daemonset"
	kubernetesDeployment                   = "kubernetes-deployment"
	linuxArm64                             = "linux-arm64"
	linuxAmd64                             = "linux-amd64"
	linuxArm                               = "linux-arm"
	darwinArm64                            = "darwin-arm64"
	darwinAmd64                            = "darwin-amd64"
	openshiftDaemonset                     = "openshift-daemonset"
	windowsAmd64                           = "windows-amd64"
)

// ErrConfigurationNotSet is returned when a platform requires an initial configuration but one is not set
var ErrConfigurationNotSet = errors.New("configuration must be set for kubernetes installation")

var platformAliases = map[string]supportedPlatform{
	// aliases
	"windows":               windowsAmd64,
	"linux":                 linuxAmd64,
	"darwin":                darwinArm64,
	"macos":                 darwinArm64,
	"macos-arm64":           darwinArm64,
	"macos-amd64":           darwinAmd64,
	"kubernetes-daemonset":  kubernetesDaemonset,
	"kubernetes-deployment": kubernetesDeployment,
	"openshift-daemonset":   openshiftDaemonset,

	// include supportedPlatform here for validation
	"linux-arm64":   linuxArm64,
	"linux-amd64":   linuxAmd64,
	"linux-arm":     linuxArm,
	"darwin-arm64":  darwinArm64,
	"darwin-amd64":  darwinAmd64,
	"windows-amd64": windowsAmd64,
}
```

jsirianni (approved):
Working great for me. Left one comment.

antonblock (approved):

model/resource_type.go
dsvanlani: I think we should, but I'm in favor of doing it in another PR

PR #1018, opened 2023-05-02
refactor: Add new prerun for init

Mrod1598:
### Proposed Change

- Put config validation in separate function

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


cli/loader.go
```diff
@@ -27,6 +27,9 @@ type Loader interface {
 
 	// LoadProfile loads the profile with the given name
 	LoadProfile(ctx context.Context, name string) error
+
+	// ValidateConfig validates the configuration.
+	ValidateConfig(_ context.Context) error
```
antonblock, line 32:
```suggestion
	ValidateConfig(ctx context.Context) error
```
Should probably stay consistent with the other methods

cli/loader.go
```diff
@@ -37,9 +40,50 @@ func AddValidationPrerun(cmd *cobra.Command, loader Loader) *cobra.Command {
...
+		cmd.PersistentPreRunE = func(cmd *cobra.Command, args []string) error {
+			if err := configLoadPreRun(cmd, args); err != nil {
+				return err
+			}
+
+			return oldPreRun(cmd, args)
 		}
+	} else {
+		cmd.PersistentPreRunE = configLoadPreRun
+	}
```
cpheps, line 68: What's the case we need this wrapping?

Mrod1598: the init expects the config to be loaded but not validated for using the defaults in the config already

cpheps: Ohh we don't have a case for this now but it's future proofing I think. I also didn't notice this is at the bottom of the file.  Is there someway to cleanup/reuse that?

cli/loader.go
```diff
@@ -37,9 +40,50 @@ func AddValidationPrerun(cmd *cobra.Command, loader Loader) *cobra.Command {
...
-			return loader.LoadConfig(ctx, configArg)
+			err := loader.LoadConfig(ctx, configArg)
+			if err != nil {
+				return err
+			}
+		}
+		err := loader.LoadProfile(ctx, profileArg)
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
Mrod1598: condensed it a good bit

antonblock (approved):

cli/loader.go
```diff
@@ -27,10 +27,46 @@ type Loader interface {
 
 	// LoadProfile loads the profile with the given name
 	LoadProfile(ctx context.Context, name string) error
+
+	// ValidateConfig validates the configuration.
+	ValidateConfig(ctx context.Context) error
+}
+
+// PreRun is a function that runs before the command
+type PreRun func(cmd *cobra.Command, args []string) error
```
cpheps, line 36: nit: This probably doesn't need to be exported.

cli/loader.go
```diff
@@ -27,10 +27,46 @@ type Loader interface {
...
+
+// AddPrerunToExistingCmd adds a pre run to an existing command
+func AddPrerunToExistingCmd(cmd *cobra.Command, new PreRun) *cobra.Command {
```
cpheps, line 39: nit: This probably doesn't need to be exported either.

cpheps (approved):
LGTM, few nits

PR #1019, opened 2023-05-02
feat(destination): Add AWS S3 destination

antonblock:
(Relies on collector v1.25.0 release)

### Proposed Change

- Add S3 destination type which uses the `awss3exporter` exporter in the collector
- This will require additional documentation
- I think the icon could be improved, but it's straight from their assets bundle

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


resources/destination-types/aws_s3.yaml
```diff
@@ -0,0 +1,124 @@
...
+    - openshift-deployment
+  parameters:
+    - name: telemetry_types
+      label: Telemetry Types
+      type: enums
+      description: "Select which types of telemetry to export."
+      validValues: ["Metrics", "Logs", "Traces"]
+      default: []
+      options:
+        gridColumns: 12
```
BinaryFissionGames, line 26: Do we think it makes sense to have all telemetry types enabled by default, instead of all disabled by default? I think having the telemetry types opt-out instead of opt-in is easier for the user.

antonblock: That works for me

BinaryFissionGames (approved):

cpheps (approved):

PR #1022, opened 2023-05-03
feat(UI): Weight Topology Graph Edges by data flow, animate active pipeline edges.

dsvanlani:
### Proposed Change
This change tracks the max ingestion value for configuration measurements and then weights the line appropriately.  This is a work in progress an open to feedback.

Changes:
- Add `maxMetricValue`, `MaxLogValue`, `maxTraceValue` fields to the OverviewMetrics and ConfigurationMetrics queries.
- When receiving Data from the metrics subscriptions we track them in the state of `OverviewPageContext` or `PipelineGraphContext`.
- Refactors `CustomEdge.tsx` out to a shared `GraphComponents` directory.
- Animates lines for active pipeline types.
- `CustomEdge` uses the `maxValues` prop to assign a stroke-width class.
- `updateGraphMetrics` now assigns each edge a `rawValue` to be used to determine width.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


ui/src/components/GraphComponents/GraphGradient.tsx
```diff
@@ -0,0 +1,23 @@
+export const GraphGradient: React.FC = () => {
+  return (
+    <svg width="20" height="20">
+      {/* This makes a gradient background available to the custom edges */}
+      <defs>
+        <linearGradient id="gradient" x1="0%" y1="0%" x2="100%" y2="0%">
+          <stop offset="0%" stopColor="#4abaeb" />
+          <stop offset="100%" stopColor="#5dc4b6" />
```
BinaryFissionGames, line 8: This is pretty nitpicky, but I notice hard edges where the gradient starts/ends, it makes these line artifacts appear - might be smoother to transition back to the start color? Might also make sense to double the pattern and rect width below if you decide to take this change.
```suggestion
          <stop offset="0%" stopColor="#4abaeb" />
          <stop offset="50%" stopColor="#5dc4b6" />
          <stop offset="100%" stopColor="#4abaeb" />
```

ui/src/components/GraphComponents/custom-edge.module.scss
```diff
@@ -0,0 +1,53 @@
...
+.metric {
+  font-size: 10px;
+}
+
+.overview-metric {
+  font-size: 12px;
+}
+
+.highlight {
+  filter: drop-shadow(2px 2px 2px colors.$summer-sky);
```
BinaryFissionGames, line 12: How do you feel about the drop shadow color here? I'm thinking it maybe should be darker (with the wide lines, it feels like it's maybe too similar to the line color), but also this color is probably fine, just want your thoughts.

antonblock (approved):

ui/src/components/GraphComponents/custom-edge.module.scss
dsvanlani: Yeah I'm not thrilled with the shadow.  I'll try the darker blue color for more contrast.

ui/src/components/GraphComponents/GraphGradient.tsx
dsvanlani: Good idea

ui/src/components/GraphComponents/custom-edge.module.scss
dsvanlani: image
 
I think the active-blue looks a little better.

BinaryFissionGames (approved):

PR #1033, opened 2023-05-04
feat: Add Datadog destination

jmwilliams89:
### Proposed Change
- Added a Datadog destination that uses the datadog exporter.
  - For metrics, `resource_attributes_as_tags` has been hardcoded as true. Without this, resource attributes would be dropped. I took an opinionated stance on this, because resource attributes are often valuable for dimensioning metrics and providing context.
  - For logs, I discovered that the datadog exporter does not support structured log bodies. Instead, there's an expectation for parsed data to be in attributes instead. The body of the log is converted to a string and saved as a `message` field. This will lead to a less than ideal experience in datadog. Unfortunately, this issue is outside the scope of this effort. It will be addressed at a future date.
  - Also for logs, it appears that resource attributes are dropped altogether. I plan on submitting an issue upstream.

#### Example Icon
image

#### Example Config
image

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


resources/destination-types/datadog.yaml
```diff
@@ -0,0 +1,193 @@
+apiVersion: bindplane.observiq.com/v1
```
jsirianni, line 1: We should add a version and probably a ticket to add versions to all resources.

resources/destination-types/datadog.yaml
```diff
@@ -0,0 +1,193 @@
...
+      required: false
+      relevantIf:
+        - name: sending_queue_enabled
+          operator: equals
+          value: true
+        - name: persistent_queue_enabled
+          operator: equals
+          value: true
+
+  logs+metrics+traces:
```
jsirianni, line 164: Do we want Datadog to be all or nothing, or should we have boolean toggles for enabling / disabling each pipeline independently?

resources/destination-types/datadog.yaml
```diff
@@ -0,0 +1,193 @@
...
+        - US1-FED
+        - AP1
+      default: US1
+      documentation:
+        - text: Read more
+          url: https://docs.datadoghq.com/getting_started/site/
+
+    - name: api_key
+      label: API Key
+      description: The API Key used for authentication.
```
jsirianni, line 30: We should mention the required permissions / scopes, if it is brief. Otherwise definitely mention in the readme.io docs.

andykellr:
My only concern about `resource_attributes_as_tags` is that it could increase cardinality substantially which may add to cost. Should we put it under an advanced section?

resources/destination-types/datadog.yaml
jsirianni: Nevermind, I was able to get this working without further configuration / permissions.

resources/destination-types/datadog.yaml
andykellr: None of our other destinations have toggles like this, but maybe we should add a processor that allows a telemetry type to be dropped completely that could be added before a destination.

jsirianni: Coralogix does but most do not https://github.com/observIQ/bindplane-op/blob/release/v1.15.0/resources/destination-types/coralogix.yaml

resources/destination-types/datadog.yaml
```diff
@@ -0,0 +1,193 @@
...
+      description: The Datadog site to send telemetry to.
+      required: true
+      type: enum
+      validValues:
+        - US1
+        - US3
+        - US5
+        - EU1
+        - US1-FED
+        - AP1
```
andykellr, line 22: Pretty minor, but should we use the same order as the Datadog site? Maybe you pulled these values from a different page that had a different order?

Screenshot 2023-05-04 at 4 23 09 PM

jsirianni:
> My only concern about `resource_attributes_as_tags` is that it could increase cardinality substantially which may add to cost. Should we put it under an advanced section?

I think the main issue is that OTLP telemetry generally places identifiable information under resource, like `host.name`. Without resource attributes, you loose the ability to differentiate between multiple systems outputting the same metrics.

We could expose the option, but I cannot think of a situation where someone would want to disable it. They can always use processors to remove unnecessary resources, like `os.type` and `host.id`.

resources/destination-types/datadog.yaml
andykellr: Also, EU or EU1?

jmwilliams89: Good catch on EU1. That was a github copilot error. Along with AP1 and US1-FED being switch in order.

resources/destination-types/datadog.yaml
jmwilliams89: I avoided the toggles because Coralogix was the only one I could find following that example. Since the Google destination has been our most popular, I tried to follow that convention instead.

resources/destination-types/datadog.yaml
jmwilliams89: I ran into this early on when I mistook application keys for an api key. It looks like only application keys have refined scope and none of them correspond to metric/log write permissions.

resources/destination-types/datadog.yaml
jmwilliams89: Actually, nevermind. I don't think there's an issue here after all. I pulled the supported list from this documentation: https://docs.datadoghq.com/getting_started/site/. The destination is currently following the order as presented in that table and it looks like it's EU1 here.

resources/destination-types/datadog.yaml
jmwilliams89: I think this should be added with all the other resources once versions/rollouts are implemented. I want to keep this effort in scope and then focus on versions separately. Once versions are in though, I do think it should be a prerequisite when adding future destinations.

jsirianni (approved):

PR #1035, opened 2023-05-04
feat: Strict unmarshal resources when applying them

BinaryFissionGames:
### Proposed Change
* When applying resources, error if there are extra keys.
  * This helps catch typos in resources, instead of the mistyped key being silently dropped.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


resources/source-types/common_event_format.yaml
```diff
@@ -28,7 +28,6 @@ spec:
       options:
         gridColumns: 12
       default: []
-      advanced_config: true
```
BinaryFissionGames, line 31: This isn't advanced in the file log, so I think we should be OK with the default for `advancedConfig` (false).

andykellr (approved):
Awesome

dsvanlani (approved):

PR #1040, opened 2023-05-05
chore(ci): Speed up PR CI 

jsirianni:
### Proposed Change

Reduced CI time by around half? Build times are somewhat inconsistent. 
- new time: 11 - 14 minutes
- old time: 18 - 26 minutes

Made a few changes to speed CI up
- Moved UI codecov to `ui-test` stage to avoid running UI tests twice
  - Was previously in the codecov stage, which has been removed
- Moved Go codecov to `go-test` stage
  - Was previously in the codecov stage, which has been removed
  - Makefile was updated to not run integration tests with codecov, because they require Goreleaser to have finished.
- Added `--split` flag to goreleaser, which allows us to build for linux only (still builds packages and container images)
- Removed dockerhub and google container repository login steps and associated pushes
  - We can rely on PR builds that get pushed to ghcr.io
- Added `test-integration` target to Makefile for running integration tests **after** Goreleaser finishes.
  - This allowed the removal of the codecov stage
- Removed package uninstall (no need to uninstall it)
- Removed client package install and uninstall, it is reliable enough and we don't actually test it. We also may be removing bindplanectl entirely.

Changes did not result in less test coverage. Codecov will report less now that it is not handling integration tests.

We can save more time by looking into the following
- Speeding up goreleaser pre hook (ui dependency and build)
- Removing bindplanectl
- Updating integration tests to run a server in code instead of relying on docker.
  - Will remove the need to pull previously pushed images
  - Will remove the need to run integration tests after Goreleaser builds

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


Makefile
```diff
@@ -61,13 +61,23 @@ ui/node_modules: ui/package.json ui/package-lock.json
 dev: ui/node_modules prep
 	./ui/node_modules/.bin/concurrently -c blue,magenta,cyan -n sv,ui,gq "go run ./cmd/bindplane/main.go serve --force-console-color --env development" "cd ui && npm start" "cd ui && npm run generate:watch"
 
-.PHONY: test # runs go test for server tests
+.PHONY: test
```
cpheps, line 64: Does this need to run with the `-cover` flag if we aren't using the cover output?

cpheps (approved):

BinaryFissionGames (approved):

Makefile
jsirianni: Nope, removed

PR #1051, opened 2023-05-09
Add Rollouts models

Mrod1598:
### Proposed Change
- Add rollouts models

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


model/resource_test.go
```diff
@@ -23,8 +25,54 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v3"
+)
+
+var (
+	// testConfigCommonConfig = config.Config{
```
dsvanlani, line 32: Is this still neccessary?

dsvanlani (approved):
Unsure about that comment - was this in the Rollouts upstream branch?

model/configuration.go
```diff
@@ -106,12 +115,232 @@ type ConfigurationSpec struct {
...
+
+func (s RolloutStatus) String() string {
+	str, ok := rolloutStatusMap[s]
+	if !ok {
+		return "none"
+	}
+	return str
+}
+
+// TODO make these strings nice
```
cpheps, line 195: We should ask if this TODO is still valid or if we're happy with these

cpheps (approved):

model/configuration.go
Mrod1598: tossed a comment on the original pr

PR #1064, opened 2023-05-11
feat(rollouts): Ported server startup logic

cpheps:
### Proposed Change
- Ported scheduler startup into server
- Refactored scheduler to be a bit more robust in start/stop
- Removed SystemStore as all store implementations we have have to be SystemStores. It was adding extra type casting when we wanted to use it.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


Mrod1598 (approved):

cli/commands/serve/server.go
```diff
@@ -129,23 +129,11 @@ func (s *defaultServer) Serve(ctx context.Context) error {
...
-	if err = s.tracer.Start(ctx); err != nil {
-		s.logger.Warn("Failed to start tracer", zap.Error(err))
-	}
-
-	s.stopQueue.Add(
-		func(stopCtx context.Context) error {
-			return s.tracer.Shutdown(stopCtx)
-		},
-	)
+	s.startScheduler(ctx)
```
Mrod1598, line 132: does it matter what order these are added to the stop queue?

cpheps: Not right now it doesn't. The stop queue will execute in FIFO order. The original design was a stack so this would work like defers but we haven't needed that functionality yet.

PR #1082, opened 2023-05-16
Rollouts/add configurable scheduler

Mrod1598:
### Proposed Change

- Make scheduler configurable

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


config/config.go
```diff
@@ -53,6 +54,8 @@ type Config struct {
 	// any other services. It will still allow agents to connect and serve api requests.
 	Offline bool `mapstructure:"offline,omitempty" yaml:"offline,omitempty"`
 
+	RolloutsInterval time.Duration `mapstructure:"rolloutsInterval" yaml:"rolloutsInterval,omitempty"`
```
cpheps, line 57: nit: needs comment

config/network.go
```diff
@@ -27,6 +28,9 @@ const (
 
 	// DefaultPort is the default port on which the server will serve.
 	DefaultPort = "3001"
+
+	// UpdateRolloutsInterval is the interval at which rollouts are updated
+	UpdateRolloutsInterval = 10 * time.Second
```
cpheps, line 33: This should be called `DefaultRolloutsInterval`. Might be worth checking this value too as I'm not sure if it changed to 5 seconds in enterprise.

server/scheduler.go
```diff
@@ -38,31 +36,36 @@ type Scheduler interface {
...
+		logger:   logger,
+		interval: interval,
 	}
 }
 
 // Start starts the scheduler
 func (s *defaultScheduler) Start(ctx context.Context) {
+	if s.interval == 0 {
+		s.interval = config.UpdateRolloutsInterval
+	}
```
cpheps, line 61: This isn't necessary. If have this restriction we should add a validation error in the config that it must be greater than 0.

cpheps (changes requested):

server/scheduler.go
cpheps: Ok, we should update the test config rather than setting a failsafe here that is only for a the test.

config/config.go
Mrod1598: Added

config/override.go
```diff
@@ -173,6 +173,7 @@ func DefaultOverrides() []*Override {
 		NewOverride("env", "env to use. One of test|development|production", EnvProduction),
 		NewOverrideWithShortHand("output", "o", "output format for client commands. One of: json|table|yaml|raw", DefaultOutput),
 		NewOverride("offline", "whether the server is in offline mode", false),
+		NewOverride("rolloutsInterval", "rate at which to update Rollouts progress", DefaultRolloutsInterval),
```
andykellr, line 176:
```suggestion
		NewOverride("rolloutsInterval", "interval between updates of rollouts", DefaultRolloutsInterval),
```

andykellr (approved):
Suggested you tweak the flag description, but looks good overall. Thanks!

cpheps (approved):

PR #1083, opened 2023-05-16
fix: Fix issue where adding the same destination twice only shows a single destination

BinaryFissionGames:
### Proposed Change
Fixes an issue where adding the same destination twice updates the layout, but only one destination node appears.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


ui/src/pages/overview/nodes/OverviewDestinationNode.tsx
```diff
@@ -32,16 +31,12 @@ export function OverviewDestinationNode(params: {
     <div
       onMouseEnter={() => setHoveredNodeAndEdgeSet(connectedNodesAndEdges)}
       onMouseLeave={() => setHoveredNodeAndEdgeSet([])}
-    >
-      {attributes.isInline ? (
-        <InlineDestinationCard id={id.replace("destination/", "")} key={id} />
-      ) : (
-        <OverviewDestinationCard
```
BinaryFissionGames, line 39: I removed to inline card here, the backend ignores inline destinations and does not send inline nodes to the frontend.

model/configuration.go
```diff
@@ -707,16 +707,19 @@ func (c *Configuration) Graph(ctx context.Context, store ResourceStore) (*graph.
 	}
 
 	for i, destination := range c.Spec.Destinations {
-		destinationName := destination.localName(KindDestination, i)
-		usage := pipelineUsage.destinations.usage(destinationName)
+		// We don't use the name, because the same destination may be used multiple times.
+		// Using the index guarantees uniqueness.
+		destinationSlug := fmt.Sprintf("destination-%d", i)
+		usage := pipelineUsage.destinations.usage(destinationSlug)
```
dsvanlani, line 713:
```suggestion
                 destinationName := destination.localName(KindDestination, i)
		usage := pipelineUsage.destinations.usage(destinationName)
```

dsvanlani (changes requested):
Working well, but I think we need the destinationName to determine correct usage.

model/configuration.go
BinaryFissionGames: Good catch, done!

dsvanlani (approved):

antonblock (approved):

PR #1103, opened 2023-05-22
refactor: Add server init for remote URL

cpheps:
### Proposed Change
Added a server init question for the user to optionally specify the `remoteURL`

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


cli/commands/initialize/initializer.go
```diff
@@ -102,8 +102,11 @@ func (i *defaultInitializer) InitializeServer(ctx context.Context) error {
 	newCfg.Auth.SecretKey = answers.SecretKey
 	newCfg.Auth.SessionSecret = answers.SessionsSecret
 
-	// Set the remote URL to the hostname if we bind to all ports
-	if answers.Host == "0.0.0.0" {
+	// Logic for setting remoteURL
+	switch {
+	case answers.RemoteURL != "": // If the remote URL is set use that
+		newCfg.Network.RemoteURL = answers.RemoteURL
+	case answers.Host == "0.0.0.0": // Set the remote URL to the hostname if we bind to all ports
```
jsirianni, line 109:
```suggestion
	case answers.Host == "0.0.0.0": // Set the remote URL to the hostname if we bind to all interfaces
```

cli/commands/initialize/initializer.go
```diff
@@ -102,8 +102,11 @@ func (i *defaultInitializer) InitializeServer(ctx context.Context) error {
...
 		hostname, _ := os.Hostname()
```
jsirianni, line 110: iirc we skip checking the error because we only expect a failure on weird distributions. Might be worth a comment.

jsirianni (commented):
Before, our default value for the prompt was dynamic depending on the value of `host`.

if host != "0.0.0.0", remote url would use the value of host.

if host == "0.0.0.0", remote url would use the value of os.Hostname().

http://127.0.0.1:3001 is fine as a default value if remote url is not set, but I think we should prefer hostname.

This is what I see
```
The full HTTP URL used for communication between client and server.
Use the IP address or hostname of the server, starting with http:// for plain text or https:// for TLS.
? Remote URL [? for help] (http://127.0.0.1:3001)
```

Mrod1598 (approved):

jmwilliams89 (approved):

jsirianni:
> Before, our default value for the prompt was dynamic depending on the value of `host`.
> 
> if host != "0.0.0.0", remote url would use the value of host.
> 
> if host == "0.0.0.0", remote url would use the value of os.Hostname().
> 
> http://127.0.0.1:3001 is fine as a default value if remote url is not set, but I think we should prefer hostname.
> 
> This is what I see
> 
> ```
> The full HTTP URL used for communication between client and server.
> Use the IP address or hostname of the server, starting with http:// for plain text or https:// for TLS.
> ? Remote URL [? for help] (http://127.0.0.1:3001)
> ```

Im still seeing this. If we need to pick one option, I think we should suggest `os.Hostname()` if the current value is `http://127.0.0.1:3001`. I suspect I keep seeing 127.0.0.1 because it is the default value baked into the default config.

cpheps:
Maybe default to hostname if the value isn't set or if it's `0.0.0.0` or `127.0.0.1`?

jsirianni:
> Maybe default to hostname if the value isn't set or if it's `0.0.0.0` or `127.0.0.1`?

I think we should always suggest hostname unless host is set to a specific IP address. The problem with 127.0.0.1 is that it is in the configuration file by default. I think we could comment remoteURL out in the default config, so that the suggestion is always hostname, unless the user has previously configured 127.0.0.1.

I like suggesting hostname because it indicates to the user that remoteURL is suitable for a dns name.

cpheps:
@jsirianni We had a dynamic question pattern in Enterprise refactor that I ported over here. I also commented out the default remoteURL in the config as it was a weird case that would have required a special case to account for. Since remoteURL is optional it makes no difference in actual usage.

jmwilliams89 (approved):

PR #1137, opened 2023-05-30
feat(destination-type): Support custom Coralogix domain

antonblock:
### Proposed Change

- Support custom domain in addition to presets
- Add AWS PrivateLink domains
- Make `subsystem_name` required
- Make the resource attribute options not advanced

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


resources/destination-types/coralogix.yaml
```diff
@@ -40,58 +50,37 @@ spec:
...
+        - name: region
+          operator: equals
+          value: "Custom Domain"
 
     - name: resource_attributes
       label: Resource Attributes
       type: bool
       default: false
       options:
         sectionHeader: true
```
BinaryFissionGames, line 77: I think it would be helpful to have a description on this parameter.

resources/destination-types/coralogix.yaml
```diff
@@ -40,58 +50,37 @@ spec:
         - APAC2
         - EUROPE1
         - EUROPE2
+        - (AWS PrivateLink) USA1
+        - (AWS PrivateLink) APAC1
+        - (AWS PrivateLink) APAC2
+        - (AWS PrivateLink) EUROPE1
+        - (AWS PrivateLink) EUROPE2
```
BinaryFissionGames, line 57: Is it obvious what these values map to for customers? e.g. that APAC1 maps to ap-south1, but AP2 maps to ap-southeast-1?

Not sure what it looks like on the coralogix side. I know the endpoint docs used to use these, but I actually don't see references to the endpoints like these anymore.

resources/destination-types/coralogix.yaml
```diff
@@ -292,6 +320,11 @@ spec:
           {{ bpRenderOtelRetryOnFailureConfig .retry_on_failure_enabled .retry_on_failure_initial_interval .retry_on_failure_max_interval .retry_on_failure_max_elapsed_time }}
           {{ bpRenderOtelSendingQueueConfig .sending_queue_enabled .persistent_queue_enabled (bpComponentID "file_storage") .sending_queue_num_consumers .sending_queue_queue_size }}
       {{ end }}
+    processors: |
+      - resourcedetection:
+          detectors: ["system", "env"]
+          system:
+            hostname_sources: ["os"]
```
BinaryFissionGames, line 327: nit: Is the "env" detector useful here? Not against having it, just curious if we're expecting it to get much use, since it requires the `OTEL_RESOURCE_ATTRIBUTES` env to be set.
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/resourcedetectionprocessor/README.md#environment-variable

resources/destination-types/coralogix.yaml
```diff
@@ -292,6 +320,11 @@ spec:
           {{ bpRenderOtelRetryOnFailureConfig .retry_on_failure_enabled .retry_on_failure_initial_interval .retry_on_failure_max_interval .retry_on_failure_max_elapsed_time }}
           {{ bpRenderOtelSendingQueueConfig .sending_queue_enabled .persistent_queue_enabled (bpComponentID "file_storage") .sending_queue_num_consumers .sending_queue_queue_size }}
       {{ end }}
+    processors: |
+      - resourcedetection:
```
BinaryFissionGames, line 324: nit: Should we set `override` to false on these? Otherwise if this were on an aggregator node, all the host info would be for the aggregator, instead of the actual collectors collecting.

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/resourcedetectionprocessor/README.md#configuration

resources/destination-types/coralogix.yaml
BinaryFissionGames: Oh, I was looking at: 
https://coralogix.com/docs/coralogix-amazon-web-services-aws-privatelink-endpoints/

But it's actually listed like this here:
https://coralogix.com/docs/coralogix-domain/

This is fine as is, I was just looking at the wrong doc :)

resources/destination-types/coralogix.yaml
antonblock: That's a good point, we don't add that detector in other places

docs/swagger/swagger.json
```diff
@@ -1145,6 +1145,242 @@
                 }
             }
         },
+        "/rollouts": {
```
Mrod1598, line 1148: was it just not generated before?

antonblock: It was not

Mrod1598 (approved):

PR #1191, opened 2023-06-14
feat(added read more option for int fields)

BominRahmani:
### Proposed Change

Added an option to see the read more option with a link on int params. Previously would not work with them.
image

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


ui/src/components/ResourceConfigForm/ParameterInput/IntParamInput.tsx
```diff
@@ -1,4 +1,4 @@
-import { FormHelperText, TextField } from "@mui/material";
+import { FormHelperText, TextField, Stack } from "@mui/material";
```
antonblock, line 1:
```suggestion
import { FormHelperText, Stack, TextField } from "@mui/material";
```

ui/src/components/ResourceConfigForm/ParameterInput/IntParamInput.tsx
```diff
@@ -45,6 +45,15 @@ const IntParamInputComponent: React.FC<ParamInputProps<number>> = ({
               {errors[definition.name]}
             </FormHelperText>
           )}
+          {definition.documentation && (
+            <Stack component={"span"}>
+              {definition.documentation.map((d) => (
+                <a href={d.url} rel="noreferrer" target="_blank">
```
antonblock, line 51:
```suggestion
                <a href={d.url} rel="noreferrer" target="_blank" key={d.url}>
```
If each item doesn't have a key we get a console error (https://github.com/observIQ/bindplane-op/pull/1175)

antonblock (changes requested):

antonblock (approved):

cpheps (approved):

PR #1193, opened 2023-06-15
feat: Microsoft 365 source type

kuiperda:
### Proposed Change
Adding a new BindPlane OP source type for the Microsoft 365 receiver, in our distro of the collector. Testing can be done by building our collector from `release/v1.27.0` or later. 

Note that the receiver only scrapes once every 24 hours, so we have an oddly large collection interval set as the default. This was documented in the description for that parameter but that made CI fail. We should find a different way to document this. 

##### Checklist
- [x] Changes are tested
- [x] CI has passed


kuiperda (commented):
Changes from feedback: 
- Hid Client Secret 
- Updated metrics collection interval units, description, and default value
- Removed advanced parameters for logs
- Hard coded storage offset directory in config

BinaryFissionGames (approved):

Mrod1598 (approved):

PR #1213, opened 2023-06-19
Fix ProcessorType => Configuration update dependency

andykellr:
### Proposed Change
When a ProcessorType is updated, check Processors defined within Sources and Destinations that are defined within a Configuration to see if the Configuration needs to be updated.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


store/basic_event_updates.go
```diff
@@ -419,6 +422,9 @@ func (u *EventUpdates) AffectsConfiguration(configuration *model.Configuration)
 		if u.DestinationTypesField.ContainsKey(destination.Type) {
 			return true
 		}
+		if u.AffectsResourceProcessors(destination.Processors) {
```
jmwilliams89, line 425: Can we condense these conditionals into a single conditional using `||`. Maybe each can be on its own line to make it more readable.

Mrod1598 (approved):

jmwilliams89 (approved):

PR #1217, opened 2023-06-20
fix(UI): Table delete buttons

dsvanlani:
### Proposed Change
We want to change the primary action to `Delete` instead of `Install` or `New` for the agents and configurations table respectively.

Changes:
- Agents Table: Replace Install button with Delete button when deletable agents are selected.
- Configs Table:
    - Replace New button with Delete button when configurations are selected.
    - add an `enableNew` prop to keep using this table on the overview page.
-  Destinations Table: Float the delete button right to stay consistent with other pages.

### Screenshots

image

image
image

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


shazlehu (approved):

BinaryFissionGames (commented):
I'm not able to delete destinations, is that related to this PR or is that a regression elsewhere?

BinaryFissionGames (approved):

PR #1225, opened 2023-06-22
fix: Overview top three ignores zombie agents

shazlehu:
### Proposed Change

The Top 3 button on the Overview page wasn't working as expected because the metrics included configurations that had zero agents but were still reporting measurements. This fixes the resolver to ignore those configs and the the measurements on any destinations that are shared between a deployed config and a zombie config.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


andykellr (approved):

dsvanlani (approved):
Just a change to the request, looks good.

PR #1227, opened 2023-06-23
feat: Expose extra options for the filelog source

BinaryFissionGames:
### Proposed Change
* Expose the fingerprinting size
  * This is really important to expose, some logs have long headers and could surpass the default (1MiB)
* Add ability to disable offset storage
  * In all production cases, you'll want this enabled. However, I've seen the field team having to blow away the offset storage constantly when testing changes. This makes things a little easier when you want to keep re-ingesting the same files with small tweaks to the config (having start_at = beginning, ).
* Add in the retry config
  * This is disabled by default, but it allows the filelog receiver to properly handle backpressure. There are still issues in the collector that make it less useful than it could be (biggest one being the batch processor, this issue is a good summary of problems with the collector right now: https://github.com/open-telemetry/opentelemetry-collector/issues/7460)
* Use "quote" for regexes
  * Had a customer use a regex with the single quote (`'`) in the regex, which failed to render.
##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


resources/source-types/file_log.yaml
```diff
@@ -289,13 +289,82 @@ spec:
...
+    # Retry on failure
+    - name: retry_on_failure_enabled
+      label: Enable Retry on Failure
+      description: Attempt to resend telemetry data that has failed to be transmitted to the destination.
+      type: bool
+      documentation:
+        - text: Read More
+          url: https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/README.md
+      # For the filelog receiver, the default is false. Enabling this enables backpressure, but the batch processor is still
+      # an issue: https://github.com/open-telemetry/opentelemetry-collector/issues/7460
```
jsirianni, line 316: If I understand the linked issue, the use of batch processor will negate filelog retry? If so, we should probably have a ticket to pull the batch processor out of our destinations. We do have a batch processor available for situations that require it, and it is probably more approaite than keeping them baked into our resource types.

resources/source-types/file_log.yaml
```diff
@@ -289,13 +289,82 @@ spec:
...
       advancedConfig: true
 
+    - name: fingerprint_size
+      label: Fingerprint size
+      description: The size of the file fingerprint used to distinguish unique files.
+      type: string
+      default: 1MiB
+      advancedConfig: true
+
+    - name: enable_offset_storage
```
jsirianni, line 299: I think the new offset options should be moved near the `offset_storage_dir`. And `offset_storage_dir` should probably depend on this option being set to true.

resources/source-types/file_log.yaml
```diff
@@ -289,13 +289,82 @@ spec:
...
+      default: 1
+      advancedConfig: true
+      required: false
+      relevantIf:
+        - name: retry_on_failure_enabled
+          operator: equals
+          value: true
+
+    - name: retry_on_failure_max_interval
+      label: Max interval
```
jmwilliams89, line 336: Should we include the word retry in some of these parameter labels, so it's obvious before reading the tooltip what the parameter corresponds to?

BinaryFissionGames: Could we take that as a separate polish ticket? This is copy and pasted from all the other retry on failure sections, so this is consistent with other retry setting sections. Personally, I don't think it's unclear when you look at it in the UI, but maybe it is inconsistent with how we use these header fields elsewhere:

Screen Shot 2023-06-26 at 12 15 34 PM

resources/source-types/file_log.yaml
BinaryFissionGames: Yeah, really the batch processor just needs to be fixed upstream to have the proper behavior. But it is a pretty big issue right now, because it means if your queue fills up with the batch processor in the pipeline, you'll just drop telemetry instead of applying back pressure.

BinaryFissionGames: That being said, we do disable this retry behavior by default since that's the way it is upstream. It just seems like an important option to have.

jsirianni: My worry is that a user might enable this feature, and not see a positive impact because we forced them to use batching (embedded in the destination).

To be clear, I think this PR is fine. Im wondering if we should follow up with embedded batch processor removal. @jmwilliams89 , thoughts?

Example: https://github.com/observIQ/bindplane-op/blob/release/v1.18.0/resources/destination-types/grafana_cloud.yaml#L194

jsirianni (approved):

resources/source-types/file_log.yaml
jmwilliams89: Yeah that's fine. Based on that screenshot, I think the checkbox actually makes it pretty obvious, so probably no need to address this now.

jmwilliams89 (approved):

resources/source-types/file_log.yaml
jsirianni: Depends on the volume I suppose. It would be negative for high volume configs.

So perhaps we keep our embedded batch processors with the understanding that they will conflict with this new feature until they are fixed upstream

PR #1228, opened 2023-06-23
fix: Source icon updates

kuiperda:
### Proposed Change
Changes:
- Splunk logo replaced
- Journald logo replaced
- Cisco Meraki source icon changed to generic Cisco icon (and Meraki icon removed)
- File, Custom, CEV, CSV, WC3 logos standardized (font is Nunito Sans 7pt Condensed - Bold Condensed)

Other:
- Drop shadow removed from topology graph lines (Mike requested this)
- Link BPOP header to Overview page instead of Agents
- Change code block in agent installation page to match platform and configuration width and spacing

##### Checklist
- [x] Changes are tested
- [x] CI has passed


ui/public/icons/destinations/custom.svg
dsvanlani, line 1: So I see that you have the font set to Nunito - but its not picking it up in the browser.  In general setting the font on text in SVG is pretty inconsistent across browser and definitely won't work for our customers running in `offline` environments.

The font for custom also seems pretty cramped.

Can we export these SVGs using `paths` instead of `text` tags?  We can see if it still looks cramped after using the correct font.

image

dsvanlani (changes requested):

ui/public/icons/sources/journald.svg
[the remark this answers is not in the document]
kuiperda: Ended up changing the journald icon to a generic notebook/journal after discussing with Josh.

ui/public/icons/destinations/custom.svg
kuiperda: Did this and made the text a bit smaller for the custom icon

dsvanlani (approved):

antonblock (approved):

PR #1238, opened 2023-06-26
feat: Add Group by Attributes processor

BinaryFissionGames:
### Proposed Change
* Adds a processor-type that exposes the Group by Attributes processor in the collector. 
  * https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/groupbyattrsprocessor

This processor allows log/datapoint/span level attributes to be correctly hoisted up to resource attributes. This is important for situations where you want to actually split logs into different resources that are currently in the same resource. We need this for routing to separate google projects based on the `gcp.project.id` resource attribute.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


jsirianni (commented):
The doc for this processor recommends using `batch` (I assume before the groupby processor). Should we consider bundling batch into this processor, or should it just be a documentation thing?

jsirianni:
Should I expect to see collector's own log attributes grouped and placed into resource?

Screenshot 2023-06-27 at 9 34 24 AM

BinaryFissionGames:
@jsirianni You should see whatever attribute keys you specify hoisted up to resource:
Screen Shot 2023-06-27 at 9 56 41 AM
Screen Shot 2023-06-27 at 9 59 46 AM

If you don't specify any keys, Resources with the same attributes are combined into one (this is less transparent to users in our UI, but could reduce sent data for some destinations).

As far as packaging the batch processor in here goes, this is a useful processor without it (receivers emit in batches already), so I think we should keep them separate for the time being.

jsirianni (approved):

antonblock (approved):

PR #1239, opened 2023-06-26
feat(UI): Delete and Duplicate current version from config page

dsvanlani:
### Proposed Change
This PR adds a hamburger menu to the `ConfigDetails` component which gives the users options to delete and to duplicate the current version when available.

image

image

image

image

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


ui/src/components/ConfigurationDetails/ConfigurationDetails.tsx
```diff
@@ -1,5 +1,6 @@
 import {
   Box,
+  Button,
```
antonblock, line 3:
```suggestion
```

ui/src/components/ConfigDetailsMenu/ConfigDetailsMenu.tsx
```diff
@@ -0,0 +1,142 @@
...
+              <TrashIcon width="20px" stroke={colors.red[700]} />
+            </ListItemIcon>
+            <Typography color="error">Delete</Typography>
+          </MenuItem>
+        </MenuList>
+      </Menu>
+
+      <DuplicateConfigDialog
+        currentConfigName={asCurrentVersion(configName)}
+        open={duplicateDialogOpen}
```
antonblock, line 129:
```suggestion
        open={duplicateDialogOpen}
        onClose={() => setDuplicateDialogOpen(false)}
```
I wasn't able to close the modal without this

ui/src/components/ConfigDetailsMenu/ConfigDetailsMenu.tsx
```diff
@@ -0,0 +1,143 @@
...
+    try {
+      await deleteResources([
+        {
+          kind: ResourceKind.CONFIGURATION,
+          metadata: {
+            name: configName,
+          },
+        },
+      ]);
+      enqueueSnackbar(`Deleted config ${configName}`, { variant: "success" });
```
antonblock, line 71: Nitpick but I think we should use `configuration` in the snackbar and confirmation modals, since we're using that in the navbar and duplication modal

andykellr (changes requested):
Noticed when we were looking at it that the padding of the title bars is different. Make the top one `0 24px` and the `Topology` title bar similar.

Also got into a strange state where the duplicate dialog didn't disappear and the menu didn't change.

1) Create a duplicate
2) Delete the duplicate
3) Duplicate again

andykellr (approved):

antonblock (approved):

PR #1248, opened 2023-06-27
feat: Add EULA notice to install scripts

jmwilliams89:
### Proposed Change
- Added EULA notice to macos install script
- Added EULA notice to linux install script

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


scripts/install-macos.sh
```diff
@@ -266,6 +266,12 @@ display_results() {
     info "To initialize the server, run: $(fg_cyan "${bin_dir}/bindplane init server")$(reset)"
     decrease_indent
 
+    banner 'EULA'
```
cpheps, line 269: We can probably delete this script we don't support macos installations any longer.

Mrod1598 (approved):

cpheps (approved):

scripts/install-macos.sh
jmwilliams89: Agreed. However, I'll keep the PRs separate.

PR #1255, opened 2023-06-28
fix(ui): Center buttons in config topology

shazlehu:
### Proposed Change

Yes, it would be better to do this by first calculating the button sizes and then positioning them based on that, but we haven't yet unlocked that power in Reactflow, so let's leave it for a future request that truly demands it.

Fixes: https://app.vivifyscrum.com/boards/114813/sprint-backlog/376716/BPOP-1053

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


antonblock (commented):
I think 40 might be a better offset for the Add Source button? It's tough to be sure, it's a super small difference

shazlehu:
I did get my ruler out :smile:

antonblock (approved):

PR #1256, opened 2023-06-28
feat(destinations): Expose TLS options for Kafka OTLP

antonblock:
### Proposed Change

- Enable TLS authentication option for Kafka OTLP (`kafkareceiver`)
- https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/kafkareceiver#getting-started

See Nico for getting access to a TLS env

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


jsirianni:
Does the Kafka receive support TLS without mutual authentication (no cert / key)?

jsirianni:
> Does the Kafka receive support TLS without mutual authentication (no cert / key)?

I misunderstood that this is literally an authentication option, not kafka listening with TLS for all traffic.

resources/source-types/kafka_otlp.yaml
```diff
@@ -73,6 +72,62 @@ spec:
...
+      type: string
+      required: false
+      default: ""
+      relevantIf:
+        - name: auth_type
+          operator: equals
+          value: tls
+        - name: insecure_skip_verify
+          operator: equals
+          value: true
```
jsirianni, line 98: I think we can remove this condition for ca_file, cert_file, key_file, for two reasons:
1. I think the kafka receiver doc has things backwards. You always want to setup certificates when tls verification is enabled (not skipped).
2. I think you could end up in a situation where you configure certificates and skip verification, so exposing all four options together might make sense.

I have not actually tested this but my gut is saying the receiver doc has things backwards.

Mrod1598 (approved):

PR #1271, opened 2023-07-05
feat: Add timezone parameter to file source

BinaryFissionGames:
### Proposed Change
* Adds a timezone parameter to the file source
  * The timezone only applies if there is not an offset in the timestamp (meaning the timezone is ambiguous). 
* Factor out parsing operator into a separate helper. 

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


resources/helpers/operatorhelper/operators.go
```diff
@@ -0,0 +1,123 @@
...
+	parseTimestampFormat string,
+	epochTimestampFormat string,
+	manualTimestampFormat string,
+	parseSeverity bool,
+	severityField string,
+) string {
+	operator := map[string]any{}
+
+	switch parseFormat {
+	case "json":
```
antonblock, line 48: This makes sense given the check for `none` in the template, but I think it'd be easier to reuse this if we can add some docs about the assumptions we make for this to be called. Like `This should only be used in a template when parseFormat is set to something other than none`

resources/source-types/file_log.yaml
```diff
@@ -114,6 +114,19 @@ spec:
           operator: notEquals
           value: none
 
+    - name: timezone
+      label: Timezone
+      description: The timezone to use when parsing ambiguous timestamps.
```
antonblock, line 119:
```suggestion
      description: The timezone to use if Timestamp Format doesn't include a time zone.
```

resources/helpers/operatorhelper/operators.go
```diff
@@ -0,0 +1,123 @@
...
+	case "ISO8601":
+		timestampConfig["layout"] = "%Y-%m-%dT%H:%M:%S.%f%z"
+	case "RFC3339":
+		timestampConfig["layout"] = "%Y-%m-%dT%H:%M:%S.%f"
+	case "Manual":
+		timestampConfig["layout"] = manualFormat
+	case "Epoch":
+		timestampConfig["layout_type"] = "epoch"
+		timestampConfig["layout"] = epochFormat
+	}
```
antonblock, line 119: I also don't know how much this will be expanded, but it might make it a bit easier to put the "magic" strings in `const`s 🤷

resources/source-types/file_log.yaml
BinaryFissionGames: I do think your wording is more clear, but the problem with this wording is it's wrong when the timezone on the timestamp is something like "CST" and not an absolute offset.

For instance, the meaning of "CST" is dependent on this parameter ( it could refer to American central time, or chinese time, depending on this location: https://go.dev/play/p/mW8IqRzQ_46).

I was trying to capture that case in the wording, but it's difficult to capture succinctly, clearly, and correctly.

resources/helpers/operatorhelper/operators.go
BinaryFissionGames: I pulled the case strings out into constants.

resources/source-types/file_log.yaml
antonblock: I agree, this seems like something better documented on Readme. I think this is still clearer, just wanted to be clear that the param will be ignored if the format already has it

shazlehu (approved):

PR #1274, opened 2023-07-06
feat: Add parsing options to TCP and UDP sources

BinaryFissionGames:
### Proposed Change
* Adds options to parse TCP and UDP logs, just like the filelog receiver.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


resources/source-types/tcp.yaml
```diff
@@ -30,6 +30,168 @@ spec:
...
+      required: true
+      relevantIf:
+        - name: parse_severity
+          operator: equals
+          value: true
+        - name: parse_format
+          operator: notEquals
+          value: none
+
+    - name: parse_to
```
jsirianni, line 185: Can we use relevantIf: to hide this option when parse format is `none`?

resources/source-types/udp.yaml
```diff
@@ -24,14 +24,186 @@ spec:
...
+      required: true
+      relevantIf:
+        - name: parse_severity
+          operator: equals
+          value: true
+        - name: parse_format
+          operator: notEquals
+          value: none
+
+    - name: parse_to
```
jsirianni, line 185: Same feedback as tcp

jsirianni (commented):
Working great. Small nit

resources/source-types/tcp.yaml
BinaryFissionGames: Done for both TCP and UDP.

jsirianni (approved):

BinaryFissionGames:
@dsvanlani I'll make a separate ticket/pr for it, so we have it consistent between file, tcp, and udp.

PR #1286, opened 2023-07-10
feat: Overview auto-sort tables for "Top Three"

shazlehu:
### Proposed Change

When the user presses the "Top Three" buttons on the Overview page, the selected resources should sort to the top of the tables: https://app.vivifyscrum.com/boards/114813/sprint-backlog/376996/BPOP-1061

Also resolves: https://app.vivifyscrum.com/boards/114813/sprint-backlog/376996/BPOP-1159

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


BinaryFissionGames (approved):

ui/src/pages/destinations/DestinationsPage.tsx
```diff
@@ -167,6 +169,7 @@ export const DestinationsPageSubContent: React.FC<
         />
         <DestinationsDataGrid
           loading={data == null}
+          {...dataGridProps}
```
dsvanlani, line 172: Just making sure this is what we want, right now providing dataGridProps would not override any prop besides loading.

When I see this pattern I typically see this `{...dataGridProps}` as the last prop, so the renderer has power to override things.

I think it should either be the first prop (It only overrides props we dont define below), or the last prop (It overrides any prop).

dsvanlani (approved):

ui/src/pages/destinations/DestinationsPage.tsx
shazlehu: Good catch, sensible

PR #1295, opened 2023-07-10
fix: Seed embedded source ids

Mrod1598:
### Proposed Change

- Seed embedded sources on start up

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


cpheps (approved):
Looks good and is working

dsvanlani (approved):
This makes sense, but I personally don't believe this to be necessary.

I think that sources that were affected did not require an ID since there was not a sensitive parameter in them.  The UI Fix resolves the issue and any change to a config will correct this.  

I just think the continuation of this into multi account and multi-node deployments should be avoided.

PR #1298, opened 2023-07-11
feat: Add TLS authentication to the Kafka destination type

BinaryFissionGames:
### Proposed Change
* Adds support for TLS authentication to the kafka destination type.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


jsirianni (commented):
Seems to work well, but I am wondering if we want the two "Advanced" sections to be joined (either at the top or the bottom)

Screenshot 2023-07-11 at 9 59 35 AM

BinaryFissionGames:
@jsirianni Yeah, it DID felt a little weird to me, I'm in favor of keeping it as-is for now (it was like this previously, and the kafka (OTLP) source type is also like this), but it's worth bringing up at ticket review at the very least.

jsirianni (approved):

antonblock (approved):

PR #1300, opened 2023-07-11
Include all key fields in processor field queries

dsvanlani:
### Proposed Change
We were seeing an `invariant` error because we were not requesting all the `keyfields` for processors, which is a `ResourceConfiguration`.

To reproduce.

1. On Main add processors to a configuration - verify a white screen.
2. Check out this branch, `fix/include-keyfields-in-processor-request`, reload the page, verify it loads correctly.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


BinaryFissionGames (approved):

cpheps (approved):
Had to do a hard refresh in the browser but it fixed after that.

Mrod1598 (approved):

PR #1310, opened 2023-07-12
Remove `bindplanectl` from repository.

dsvanlani:
### Proposed Change
- Removes `bindplanectl/main.go` and `shared_commands.go`
- Removes references to `bindplanectl in documentation and comments
- Removes `bindplanectl` build from goreleaser.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


cli/commands/serve/command_test.go
```diff
@@ -0,0 +1,26 @@
+package serve
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+type mockBuilder struct{}
```
cpheps, line 10: nit: we should use mockery to generate mocks. We create an in package mock in the [transformagent](https://github.com/observIQ/bindplane-op-enterprise/blob/release/v1.21.0/internal/transformagent/manager.go#L11) package in ee

cli/commands/serve/command.go
```diff
@@ -31,6 +36,10 @@ func Command(builder Builder) *cobra.Command {
 		Short: "Starts the server",
 		Long:  `Serves websockets for agents, REST for cli, and GraphQL.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if isWindows() {
+				return errors.New("cannot start server, unsupported platform windows")
+			}
```
cpheps, line 41: I'm guessing we left out mac since we develop on macs. Should we maybe check mac if the `config.Env` is `Production`?

cli/commands/serve/command.go
```diff
@@ -58,3 +67,8 @@ func Command(builder Builder) *cobra.Command {
 
 	return cmd
 }
+
+// IsWindows returns true if the current OS is Windows
+func IsWindows() bool {
```
cpheps, line 72: nit: We can change this function so it doesn't have to be exposed but can still be use for testing.

Something like 

```go
var isWindowsFunc = func() bool {
  return runtime.GOOS == "windows"
}
```
Then just call `isWindowsFunc` in the `RunE` function. It can still be overwritten for tests.

cpheps: It also reads a bit odd that `serve.Cmd` has to take in a `serve.IsWindows` function when both are from the same package.

cli/commands/serve/command.go
dsvanlani: Thats correct, and I imagine we want to leave it for outside devs kicking the tires.

Up to you - we call it out on docs that it isn't supported, and I think it would be wild if someone tried to use a mac as their server for this kind of thing in a production environment.

I'll defer to you whether we want this check.

cli/commands/serve/command.go
dsvanlani: So I included a `SupportsServe()` method on the `Builder` interface.  I think its reasonable, and its easier to mock the builder in this case.

cli/commands/serve/command.go
cpheps: Let's bring it up in standup since this is kind of a restriction regression.

cli/commands/serve/command.go
cpheps: I like this design very clean.

cpheps (approved):

jsirianni (approved):

PR #1311, opened 2023-07-12
fix: Console error on add source

shazlehu:
### Proposed Change

We were passing a `(boolean) => void` function to the `onClick` prop of the Add Source Button, which takes `() => void`. Strangely, the state still changed but somehow caused the strange console error behavior. Not sure exactly how, but the `onClick` function was passed as a prop to the `Transition` that fades in the dialog. Assuming library updates triggered the latent problem.

The change to `ui/src/apollo-client.ts` is to fix an unrelated console warning about the Apollo cache that I encountered while working on this.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


dsvanlani (approved):
Verified there is no console error, nice fix.

antonblock (approved):

PR #1318, opened 2023-07-13
fix: ProcessorInput can have ID

dsvanlani:
### Proposed Change
We were getting an apollo 422 error when attempting to add a processor to configs with existing processors.

```json
{
    "errors": [
        {
            "message": "unknown field",
            "path": [
                "variable",
                "input",
                "processors",
                0,
                "id"
            ],
            "extensions": {
                "code": "GRAPHQL_VALIDATION_FAILED"
            }
        }
    ],
    "data": null
}
```

The fix is to include the field as optional in the `ProcessorInput` for the `updateProcessors` mutation.

### Testing

Verify Bug behavior, on `main` 
1. Add a processor to a source on a config and Save, first add works because we do not set the ID on the front end.
2. Add another processor, this fails because we are including the ID field in the first processor.  Should see a 422 error.

Verify fix on `fix/processor-input-id`
1. You should be able to add a processor to a source with existing processors.

##### Checklist
- [x] Changes are tested
- [x] CI has passed


jsirianni (approved):
Working for me :+1:

antonblock (approved):

PR #1321, opened 2023-07-13
feat(UI): Add additionalInfo to resource metadata + update Microsoft 365 source

kuiperda:
### Proposed Change
- Add an optional `additionalInfo` property to resource metadata
- Add a corresponding [mui alert](https://mui.com/material-ui/react-alert/) in TitleSection.tsx
- Update Microsoft 365 source to use an `additionalInfo`
- Update Microsoft 365 source collection interval description and default value, as well as parameter.go to pass CI. 

`additionalInfo` is intended for use on sources where we want to clearly communicate certain configuration details to customers up front. 

## Example
metadata in source.yaml that uses an `additionalInfo` property, which contains a required string `message` and an optional `documentation` array:
```
metadata:
  name: m365
  displayName: Microsoft 365
  icon: /icons/sources/microsoft_iis.svg
  description: Collect metrics and logs from Microsoft 365.
  additionalInfo:
    message: You must configure your tenant for logs, which can take 1 hour to be enabled and then up to 12 hours to become available. Metrics are reported once per day, but a report contains the metrics that were generated two days prior.
    documentation:
      - text: Read more
        url: https://docs.bindplane.observiq.com/docs/microsoft-365
```

## In the ui: 

The alert box will shrink to text size that takes less than one line. Documentation links will not wrap between lines. 
This could be expanded to allow other icon/color/title options. 

M365-source-configure

##### Checklist
- [x] Changes are tested
- [x] CI has passed


graphql/schema.graphqls
```diff
@@ -70,6 +70,12 @@ type Metadata {
   labels: Map
   version: Version!
   dateModified: Time
+  additionalInfo: AdditionalInfoBlock
+}
+
+type AdditionalInfoBlock {
+  message: String!
+  documentation: [DocumentationLink]
```
dsvanlani, line 78: I think the DocumentationLink is non nullable.

```suggestion
  documentation: [DocumentationLink!]
```

ui/src/components/DialogComponents/TitleSection.tsx
```diff
@@ -1,18 +1,27 @@
...
+  IconButton,
+  Stack,
+  Typography,
+  Alert,
+} from "@mui/material";
 import { memo } from "react";
 import { XIcon } from "../Icons";
 
 import styles from "./dialog-components.module.scss";
+import { AdditionalInfoBlock } from "../../graphql/generated";
```
dsvanlani, line 12: nit: Can you move this import to the top block?

ui/src/components/DialogComponents/TitleSection.tsx
```diff
@@ -34,6 +43,24 @@ const TitleSectionComponent: React.FC<Props> = ({
...
+              <a
+                className={styles["doclink"]}
+                href={d?.url}
+                rel="noreferrer"
+                target="_blank"
+                key={d?.url}
+              >
+                {d?.text}
+              </a>{" "}
+            </span>
```
dsvanlani, line 60: The `key` prop needs to be at the top level of the mapped component. Also a key that is possibly undefined is a bad choice, but I think making the documentationLink non nullable in the schema will fix that.

```suggestion
            <span key={d.url}>
              <a
                className={styles["doclink"]}
                href={d?.url}
                rel="noreferrer"
                target="_blank"
              >
                {d?.text}
              </a>{" "}
            </span>
```

ui/src/components/DialogComponents/TitleSection.tsx
```diff
@@ -34,6 +43,24 @@ const TitleSectionComponent: React.FC<Props> = ({
           <XIcon strokeWidth={"3"} width={"28"} />
         </IconButton>
       </Stack>
+      {additionalInfo && (
+        <Alert severity="info" className={styles["info"]}>
+          {additionalInfo.message}{" "}
```
dsvanlani, line 48: Can we put this in typography and have the links on new lines?
```suggestion
          <Typography>{additionalInfo.message}</Typography>
```

dsvanlani (changes requested):

resources/source-types/microsoft_365.yaml
```diff
@@ -5,14 +5,18 @@ metadata:
   displayName: Microsoft 365
   icon: /icons/sources/microsoft_iis.svg
   description: Collect metrics and logs from Microsoft 365.
+  additionalInfo:
+    message: You must configure your tenant for logs, which can take 1 hour to be enabled and then up to 12 hours to become available. Metrics are reported once per day, but a report contains the metrics that were generated two days prior.
+    documentation:
+      - text: Read more
```
antonblock, line 11:
```suggestion
      - text: Read More
```

kuiperda (commented):
Responded to feedback but am still adding props so the `additionalInfo` gets passed to the 'Add Source' views, not just the 'Edit Source'. Will re-ping when this is completed. Thanks for the reviews!

kuiperda (commented):
Passed props in more places so that the info alert shows up in AssistedConfig StepTwo add/edit source, and on the 'add source' in the topology of a configuration.

kuiperda (commented):
With the new change to keep the title header visible while scrolling (on sources), the info alert can take up too much room. Maybe we could try collapsing it when scrolling? Here's what it looks like currently: 

image

kuiperda (commented):
This is more of a question, if this seems right- 
- I'm noticing in the playground that sources without an additionalInfo property in their metadata do have an (empty) additionalInfo being returned. Dave mentioned that a 'Documentation' should be non-nullable but its value is actually null in these cases. 
- I'm also noticing that adding a completely empty additionalInfo to a source is allowed. Either way it  results in an empty alert being displayed if I kept {additionalInfo && (...)} as the conditional checker (instead of adding the `.message` check).

kuiperda (commented):
Last thing- should I be adding tests for this component? And I'm curious if I should document the new metadata property somewhere.

dsvanlani:
> Last thing- should I be adding tests for this component? And I'm curious if I should document the new metadata property somewhere.

Tests are good, could test the TitleSection that it renders the alert.

dsvanlani (approved):

antonblock (approved):

PR #1322, opened 2023-07-13
feat: Make the name parameter of the profile set command optional

andykellr:
### Proposed Change
If no `[name]` parameter is passed to the profile set command, use the current profile.

While I was in here, I also cleaned up some of the command descriptions and made the `--current` flag hidden since it isn't required.

```sh
Profile commands for managing BindPlane application configuration

Usage:
  bindplane profile [command]

Available Commands:
  create      Create a new profile
  current     Print the name of the current profile
  delete      Delete a profile
  get         Get details of a profile
  list        List all profiles
  set         Set details of a profile
  use         Set the current profile to the specified profile
```

To test, try

```sh
bindplane profile set --port 1
bindplane profile get  # confirm set to 1
bindplane profile set --port 3001

bindplane profile create tmp
bindplane profile set tmp --port 1
bindplane profile get tmp # confirm set to 1
bindplane profile delete tmp
```

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


cli/commands/profile/command.go
```diff
@@ -82,27 +76,29 @@ func GetCommand(builder Builder) *cobra.Command {
 		},
 	}
 
-	cmd.Flags().BoolVar(&currentFlag, "current", false, "show the settings for the current profile")
+	cmd.Flags().BoolVar(&currentFlag, "current", false, "get the settings for the current profile")
+	_ = cmd.Flags().MarkHidden("current")
```
dsvanlani, line 80: I like hiding it, I think we could just remove it too.

cli/commands/profile/command.go
```diff
@@ -266,3 +264,10 @@ func CurrentCommand(builder Builder) *cobra.Command {
 
 	return cmd
 }
+
+func profileNameArgOrCurrent(ctx context.Context, args []string, profiler Profiler) (string, error) {
+	if currentFlag || len(args) == 0 {
+		return profiler.GetCurrentProfileName(ctx)
+	}
+	return args[0], nil
```
dsvanlani, line 272: I think we should set `default` profile if there is not a current.  I blew away my profiles and got this error after `bindplane profile set --username blah`

```
Error: failed to read current: open /Users/dsvanlani/.bindplane/profiles/current: no such file or directory
exit status 1
```


```suggestion
	if len(args) > 0 {
        return args[0], nil
    }

    currentProfile, err := profiler.GetCurrentProfileName(ctx)
    if err != nil {
        return "default", nil
    }

    return currentProfile, nil
```

of course then we'd have to instruct them to `bindplane profile use default`. So I could go either way.

dsvanlani (approved):
I could go either way on the suggestion.  Working well and definitely an improvement.

cli/commands/profile/command.go
andykellr: I took a slightly different approach because I didn't want `bindplane profile get` to use `default` if there was no current profile, but it now works to delete your profiles folder and run `bindplane profile set --port 8000`.

jmwilliams89 (approved):

PR #1324, opened 2023-07-14
feat(source-type): Add TLS auth_type to kubelet

antonblock:
### Proposed Change

- Add `auth_type`, defaulting to service account
- Add fields required for TLS authentication

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


jsirianni (commented):
If skip tls verification is enabled, should we still require the ca file? (its possible the receiver does require it, not sure).

Screenshot from 2023-07-14 12-53-36

I think requiring the client key pair makes sense, because it is TLS auth after all.

jsirianni (approved):

PR #1333, opened 2023-07-17
fix: Make updates to file log sort rules

Mrod1598:
### Proposed Change

- Add link to timezone parsing
- `ascending` -> `sort direction` and move it up next to sort type
-  Move timezone next to layout 

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


jmwilliams89 (approved):

model/parameter.go
```diff
@@ -752,23 +752,15 @@ func (p ParameterDefinition) validateFileLogSortType(_ parameterFieldType, value
...
-
-		case "ascending":
-			_, ok := n.(bool)
-			if !ok {
-				return stanzaerrors.NewError("incorrect type included in "+s+" field",
-					"filelog"+s+"should be of type bool")
-			}
 		default:
 			return stanzaerrors.NewError("unexpected field "+s+" included in struct",
 				s+"should not be an included field in filelogsort")
```
andykellr, line 763: I know this isn't part of this PR, but shouldn't there be a space here?
```suggestion
				s+" should not be an included field in filelogsort")
```

Mrod1598: Updated.

PR #1334, opened 2023-07-18
fix(ui): Int param position bugfix

BominRahmani:
Fixes Int param readmore so that it appears after the description
image


shazlehu:
The PR title should start with "fix: " or "fix(ui): "

shazlehu (approved):

andykellr (approved):

PR #1348, opened 2023-07-20
feat: Allow resources to be deprecated

andykellr:
### Proposed Change
Deprecated resources will not be seeded into accounts unless they already exist. If they exist, they will be updated.

Includes the following changes:
* 4 processor-types have been deprecated and restored from git and placed in a deprecated subfolder of resources.
* Deprecated processor-types are now added to the list of folders to be seeded
* During seed, deprecated resources are only added/updated if they already exist. If you manually delete the resource and restart the server, it will not be added back. If you manually add them back using `bindplane apply -f resources/deprecated/processor-types/*.yaml` and then make some changes to those files, the changes will be applied.
* The processors UI (add and edit) has been updated to include deprecated pills to indicate deprecated resources. Also additionalInfo is now displayed for the deprecated processor-types.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


resources/resources.go
```diff
@@ -28,4 +28,5 @@ var SeedFolders = []string{
 	"source-types",
 	"processor-types",
 	"agent-versions",
+	"deprecated/processor-types",
```
dsvanlani, line 31: Should we just add `deprecated/source-types` and `deprecated/destination-types` now to avoid missing in the future? We might need to add a `.keep` file in the directories.

I don't feel strongly.

dsvanlani (approved):
Working great.

shazlehu (approved):
Works as expected

PR #1362, opened 2023-07-24
fix: Quote string inputs in metric name filter

BinaryFissionGames:
### Proposed Change
* Quote the string inputs for metric name filter, since user-provided input could cause the config to not be render-able otherwise.

##### Checklist
- [x] Changes are tested
- [x] CI has passed


jsirianni (approved):
Working great for me

jmwilliams89 (approved):

PR #1363, opened 2023-07-24
feat(cli): Support exporting resources with get command

jsirianni:
### Proposed Change

Added "export" support to the following resources
- sources
- processors
- destinations
- configurations

When the `--export` flag is passed to the `get` command, resources will be returned without select metadata such as `id`, `hash`, `date modified`, `status`, etc. Additionally, version suffix is removed from the type, such as `file:3` --> `file`.

By exporting resources, they will be in a fresh state suitable for import into a new system.

### Testing

You can test with the following script. 

1. Create one or more destinations
2. Create a configuration which uses those destinations
3. Rollout the configuration
4. Make changes to destinations and sources
5. Rollout the configuration

At this point, you have a configuration and destinations with several versions.

Export all resources:
```bash
bindplane get source -o yaml --export > source.yaml
bindplane get processor -o yaml --export > processor.yaml
bindplane get destination -o yaml --export > destination.yaml
bindplane get configuration -o yaml --export > configuration.yaml
```

You can inspect the output files. Note that they do not include version information, or generated IDs and hashes.

Delete your configurations and destinations and then import the resources from your yaml files. This will simulate a migration from one system to another.

```
bindplane apply -f source.yaml
bindplane apply -f processor.yaml
bindplane apply -f destination.yaml
bindplane apply -f configuration.yaml
```

Running a `bindplane get config <name> -o yaml` should should your configuration as version 1. You should see similar results with destinations.

I tested with Terraform, creating > 10 sources, > 10 processors, > 10 destinations and around 100 configurations. I was able to export everything and import to a fresh system.

##### Checklist
- [x] Changes are tested
- [x] CI has passed


client/bindplane.go
```diff
@@ -101,13 +101,21 @@ type BindPlane interface {
 	RawConfiguration(ctx context.Context, name string) (string, error)
 	// CopyConfig creates a deep copy of an existing resource under a new name.
 	CopyConfig(ctx context.Context, name, copyName string) error
+	// ExportConfigurations retrieves all configurations in a suitable format for export.
+	ExportConfigurations(ctx context.Context) ([]*model.Configuration, error)
+	// ExportConfiguration retrieves processor with given name in a suitable format for export.
+	ExportConfiguration(ctx context.Context, name string) (*model.Configuration, error)
```
BinaryFissionGames, line 107: Each method on the client should generally map to an API endpoint (they should be one-to-one), and these are extra logic on top of the API endpoint that's already exposed.

I think it would make more sense to move the logic for stripping metadata into the `cli/commands/get` package.

client/export.go
```diff
@@ -0,0 +1,85 @@
...
+
+import (
+	"strings"
+
+	"github.com/observiq/bindplane-op/model"
+)
+
+// exportConfiguration sanitizes a configuration for export by removing
+// fields not suitable for export, such as the ID, hash, and version.
+func exportConfiguration(c *model.Configuration) *model.Configuration {
```
BinaryFissionGames, line 25: nit: This function modifies the configuration that's passed in, it doesn't need to return a `*model.Configuration`.

client/bindplane.go
jsirianni: Makes sense, that is how I had it originally :facepalm:

cli/commands/get/export.go
```diff
@@ -0,0 +1,83 @@
...
+	m.Hash = ""
+	m.Version = 0
+	m.DateModified = nil
+	return m
+}
+
+// sanitizeTypeForExport removes the version suffix from a type
+func sanitizeTypeForExport(t string) string {
+	return strings.Split(t, ":")[0]
+}
```
BinaryFissionGames, line 83: Could we use `model.TrimVersion` here instead?

BinaryFissionGames (commented):
Seems like it's working great, just one nit.

cli/commands/get/export.go
jsirianni: Awesome, yes.

BinaryFissionGames (approved):

Mrod1598 (approved):

PR #1366, opened 2023-07-25
fix(swagger): Make swagger routes lowercase

antonblock:
### Proposed Change

Our routes are case-sensitive, so make sure the Swagger docs use the correct case

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


BinaryFissionGames (approved):

dsvanlani:
Can we change the `/Destinations` routes while we're in here?

image

antonblock:
> Can we change the `/Destinations` routes while we're in here?
> image

Thought I got all of them, good catch

jmwilliams89 (approved):

PR #1374, opened 2023-07-26
Require map parameters to specify labels for Key and Value

andykellr:
### Proposed Change
The following checks are added
* Error if `labels` field is used for non-map parameters
* Warn if `labels` field missing or doesn't specify `key` and `value` for map parameters
* Warn if `gridColumns: 12` is missing for map parameters

Tests will fail until we add map labels for a few more resources.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


dsvanlani:
Maybe for Cloudflare source:

```yaml
- name: attributes
      label: Attributes
      description: A map of Cloudflare LogPush field names to log record attributes.
      type: map
      default: {}
      advancedConfig: true
      options:
        gridColumns: 12
        labels:
          key: LogPush Field Name
          value: Attribute
```

resources/source-types/cloudflare.yaml
```diff
@@ -45,12 +45,15 @@ spec:
...
       label: Attributes
-      description: Any Cloudflare LogPush field names provided as keys will be mapped to log record attributes using the map value as the attribute name.
+      description: A map of Cloudflare LogPush field names to log record attributes.
       type: map
       default: {}
       advancedConfig: true
       options:
         gridColumns: 12
+        labels:
+          key: LogPush Field Name
```
jmwilliams89, line 55: I'd simplify this to just Field Name, since LogPush is called out in the description.

andykellr: Should we just simplify to `Field`? I considered the same issue with `Attribute` vs `Attribute Name` and we've dropped the Name in that case.

jmwilliams89: `Field` should be good in my opinion.

dsvanlani (approved):

PR #1378, opened 2023-07-26
feat: Add Store methods ReportConnectedAgents and DisconnectUnreportedAgents

andykellr:
### Proposed Change
These will be used to keep track of connected agents.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


store/boltstore_common.go
```diff
@@ -630,6 +630,89 @@ func (s *BoltstoreCore) DeleteDestinationType(ctx context.Context, name string)
 	return item, err
 }
 
+// ReportConnectedAgents sets the ReportedAt time for the specified agents to the specified time. This update should
+// not fire an update event for the agents on the Updates eventbus.
+func (s *BoltstoreCore) ReportConnectedAgents(ctx context.Context, agentIDs []string, time time.Time) error {
```
dsvanlani, line 635: I assume that this will be used by the Opamp manager - Is it likely that multiple agents will have the exact same report time?

I suppose we don't loose anything by allowing multiple agent IDs, but I was just curious if you think it would be common to include more than one.

dsvanlani (approved):

store/boltstore_common.go
andykellr: At regular intervals we will set the ReportedAt time for all connected agents to the current time. In some stores this can be done in a single operation. It will be more clear when I bring back the code that calls these methods.

store/boltstore_common.go
```diff
@@ -630,6 +630,89 @@ func (s *BoltstoreCore) DeleteDestinationType(ctx context.Context, name string)
...
+		}
+		cursor := bucket.Cursor()
+		prefix := AgentPrefix()
+
+		for k, v := cursor.Seek(prefix); k != nil && bytes.HasPrefix(k, prefix); k, v = cursor.Next() {
+			agent := &model.Agent{}
+			if err := jsoniter.Unmarshal(v, agent); err != nil {
+				// unable to unmarshal, ignore this agent
+				continue
+			}
```
BinaryFissionGames, line 686: It seems like this is a super unlikely scenario to hit, but I think there should be some log here if this happens.

store/store_test.go
```diff
@@ -4988,5 +4989,87 @@ func runTestSeedDeprecated(ctx context.Context, t *testing.T, store Store) {
...
+				if slices.Contains(test.agentIDs, agent.ID) {
+					require.Equal(t, newDate, *agent.ReportedAt)
+				} else {
+					require.Equal(t, oldDate, *agent.ReportedAt)
+				}
+			}
+
+			// disconnect unreported agents
+			err = store.DisconnectUnreportedAgents(ctx, newDate)
+
```
BinaryFissionGames, line 5060: looks like `err` here is unchecked.

store/boltstore_common.go
```diff
@@ -630,6 +630,89 @@ func (s *BoltstoreCore) DeleteDestinationType(ctx context.Context, name string)
...
+func (s *BoltstoreCore) DisconnectUnreportedAgents(ctx context.Context, since time.Time) error {
+	ctx, span := tracer.Start(ctx, "store/DisconnectUnreportedAgents")
+	defer span.End()
+
+	changes := s.CreateEventUpdate()
+
+	err := s.DB.Update(func(tx *bbolt.Tx) error {
+		bucket, err := s.AgentsBucket(ctx, tx)
+		if err != nil {
+			return err
```
BinaryFissionGames, line 676: Could we wrap the errors in this function passed to Update? I think it would make it easier to trace if we ever ended up getting one of these errors back.

store/boltstore_common.go
```diff
@@ -630,6 +630,89 @@ func (s *BoltstoreCore) DeleteDestinationType(ctx context.Context, name string)
...
+	defer span.End()
+
+	// these updates will not be reported to the eventbus
+	updates := s.CreateEventUpdate()
+
+	err := s.DB.Update(func(tx *bbolt.Tx) error {
+		bucket, err := s.AgentsBucket(ctx, tx)
+		if err != nil {
+			return err
+		}
```
BinaryFissionGames, line 646: There might be some benefit to wrapping the errors in this function too.

BinaryFissionGames (approved):

PR #1389, opened 2023-07-26
fix: Add rollup interval for 10s

shazlehu:
### Proposed Change

Stop data flowing to an agent, and this is the eventual view for 10s period (incorrect):

Screenshot 2023-07-26 at 3 24 53 PM

Versus 1m (correct):

Screenshot 2023-07-26 at 3 25 09 PM

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


jmwilliams89 (approved):

shazlehu:
This is looking more complicated. Using the shorter rollup interval makes the data a lot more erratic. In my setup with a File source that that appends one exactly 500 byte log per second, only using a one hour duration gives the right numbers (for all the different periods)

andykellr (approved):

PR #1391, opened 2023-07-26
feat: Automatically disconnect agents that have not reported recently

andykellr:
### Proposed Change
This represents the second half of the agent ReportedAt feature for updating the status of disconnected agents that have not reported recently.

The best way to test this is to comment out boltstore.go:70 `store.disconnectAllAgents(ctx)`. This will ensure that agents remain connected through a restart of the server which is necessary to detect the case where an agent stops when the server is down.

In a terminal, run the following script:

```sh
while true; do bindplane get agents; sleep 1; done
```

While watching the output of the above command, do the following:
1. Start the server
2. Connect some agents
3. Stop the server
4. Disconnect some agents
5. Start the server
6. Notice that the disconnected agents will become disconnected within 30s.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


server/manager.go
```diff
@@ -249,6 +291,52 @@ func (m *DefaultManager) AgentVersion(ctx context.Context, version string) (*mod
...
+
+func (m *DefaultManager) handleAgentDisconnect() {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	ctx, span := tracer.Start(ctx, "manager/handleAgentReporting")
+	defer span.End()
+
+	cutoff := time.Now().Add(-AgentDisconnectTTL)
+	m.Logger.Sugar().Infof("disconnecting agents not reported since %s", cutoff.Format(time.RFC3339))
```
BinaryFissionGames, line 329: Any reason to use a sugared logger here instead of the field based logging and `zap.Time`?
Also, should this be a debug log instead of info?

andykellr: Copy paste from a similar function. I will change them both.

server/manager.go
```diff
@@ -249,6 +291,52 @@ func (m *DefaultManager) AgentVersion(ctx context.Context, version string) (*mod
 
 // ----------------------------------------------------------------------
 
+func (m *DefaultManager) handleAgentReporting() {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	ctx, span := tracer.Start(ctx, "manager/handleAgentReporting")
+	defer span.End()
```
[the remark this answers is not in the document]
andykellr: We use a background so that we get an individual trace per execution of this function. If we use m.ManagerCtx, the tracing would accumulate indefinitely and never end.

Cancel does not appear to be used and I will remove it.

server/manager.go
```diff
@@ -249,16 +291,58 @@ func (m *DefaultManager) AgentVersion(ctx context.Context, version string) (*mod
...
+	}
+}
+
+func (m *DefaultManager) handleAgentDisconnect() {
+	ctx := context.Background()
+	ctx, span := tracer.Start(ctx, "manager/handleAgentReporting")
+	defer span.End()
+
+	cutoff := time.Now().Add(-AgentDisconnectTTL)
+	m.Logger.Info("disconnecting agents not reported", zap.Time("since", cutoff))
```
BinaryFissionGames, line 327: I still think this should probably be a debug log, seems like it could be a little noisy with not much benefit outside of dev.

BinaryFissionGames (approved):

dsvanlani (approved):

PR #1401, opened 2023-07-28
fix: bindplane delete should work like bindplane apply

andykellr:
### Proposed Change
The following commands will be supported:

```sh
bindplane delete -f resources/source-types/*.yaml
bindplane delete resources/source-types/*.yaml
cat resources/source-types/otlp.yaml | bindplane delete -f -
```

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


Mrod1598 (commented):
Can we get a test for deleting resources from a reader if its not overly difficult.

cpheps (commented):
LGTM Agree though can we add a tests for `DeleteResourcesFromReader`?

cpheps (approved):

PR #1415, opened 2023-07-31
Move ReportConnectedAgents to the Protocol

andykellr:
### Proposed Change
Rather than getting the list of connected agents, just let the Protocol make the call to the store.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


store/boltstore.go
```diff
@@ -67,7 +67,7 @@ func NewBoltStore(ctx context.Context, db *bbolt.DB, options Options, logger *za
 	store.BoltstoreCommon = store
 
 	// boltstore is not used for clusters, disconnect all agents
-	store.disconnectAllAgents(ctx)
+	// store.disconnectAllAgents(ctx)
```
BinaryFissionGames, line 70: nit: Should we just delete this line of code instead of just commenting it?

andykellr: Oh yeah, I'll restore that.

BinaryFissionGames (approved):

antonblock (approved):

PR #1418, opened 2023-08-01
chore: Improve `agent` package code quality

BinaryFissionGames:
### Proposed Change
* Improve code quality of agents package
* Bump test coverage up to 95.9%

Specific changes done:
* Pull out api url into a constant, and move some constants to top of `github.go`
* rename some constants to be more specific
* change receiver variable to match the first letter of the struct
* lowercase error messages
* add error wrapping where it was missing
* pull anonymous struct into a named struct, `PlatformArtifactNames`
* put a function on `PlatformArtifactNames` for formatting the download name
* move IsGitHubAgentVersion to EE
* Remove unused Offline parameter for VersionsSettings
* listen on the event bus before returning from NewVersions
  * This was mostly for making testing possible, but could avoid the very slight chance missing some agent-version events
* Capitalize log messages
* Fix nil pointer access when Offline is set to true
  * We use the Noop client in this case; Originally, no client would be set in offline mode, which lead to an immediate nil pointer panic.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


Mrod1598 (approved):

agent/github.go
```diff
@@ -30,27 +28,29 @@ var (
 	oldestReleaseDate = time.Date(2022, time.August, 1, 0, 0, 0, 0, time.UTC)
 )
 
+const githubAPIUrl = "https://api.github.com"
```
jmwilliams89, line 31: Any reason this can't be defined in the same block as the other github constants down below?

BinaryFissionGames: Nope, no reason, merged this constant with the block below it.

jmwilliams89 (approved):

PR #1423, opened 2023-08-02
fix: Use jsoniter.ConfigCompatibleWithStandardLibrary to ensure consistent marshalling

andykellr:
### Proposed Change
This also includes a small optimization to only re-index a configuration if the configuration was modified.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


BinaryFissionGames (approved):

store/boltstore.go
```diff
@@ -915,9 +915,12 @@ func editResource[R model.Resource](ctx context.Context, s BoltstoreCommon, tx *
 		return resource, false, err
 	}
 
-	if resource.GetKind() == model.KindConfiguration {
-		if err := s.ConfigurationsIndex(ctx).Upsert(resource); err != nil {
-			return resource, wasModified, err
+	// only re-index if modified
+	if wasModified {
+		if resource.GetKind() == model.KindConfiguration {
```
dsvanlani, line 920:
```suggestion
	// only re-index if modified
	if wasModified && resource.GetKind() == model.KindConfiguration {
```

andykellr: I wrote it that way initially, but preferred to wrap the block. If another resource needs to be reindexed or something it will be obvious where that goes without refactoring the conditional.

jsirianni (approved):
Excellent results

PR #1424, opened 2023-08-02
fix(panic): Check if state is nil before attempting to send message

jsirianni:
### Proposed Change

Legacy OpAMP `send` should check if the agent state is nil before attempting to send message. This resolves a potential panic.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


andykellr (changes requested):
Implementation looks good, but this should be added to internal/opamp/api.go as well.

antonblock (approved):
Looks good once non-legacy is patched

cpheps (approved):

andykellr (approved):

PR #1432, opened 2023-08-03
fix: Keep agent type for synced github resources as "observiq-otel-collector"

BinaryFissionGames:
### Proposed Change
* Because of the repo name change, the name of the agent-version resources changed (the type name was tied to the repo name). This PR makes it so that the type is fixed to the constant defined in the model package, which is what the server manager uses when looking the version up.
* Swap the order of the err and nil checks for version in the opamp package, only (nil, nil) implies that the version does not exist, not just if version is nil, because the error may be populated.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


Mrod1598 (approved):

jsirianni (approved):
Working great for me. I was able to re-produce the issue and confirm he solution.

PR #1446, opened 2023-08-07
feat: Add CSV Printing capabilities

Mrod1598:
### Proposed Change

- Add CSV Printing capabilities

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


cli/printer/csv.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// PrintResources prints a generic model that implements the model.Printable interface
+func (cp *CSVPrinter) PrintResources(list []model.Printable) {
+	if len(list) == 0 {
+		return
+	}
+
+	fieldsAndTitles := [][]string{}
+	fieldsAndTitles = append(fieldsAndTitles, list[0].PrintableFieldTitles())
```
jmwilliams89, line 57: I think we should set the titles to its own variable. That way, when we iterate later on, we can reuse that and maintain order no matter which item in the array we're looking at. Also, if somehow someone mixes printable item types, we're maybe better protected, because we're always looking for the same set of titles for every item. I'd even consider alphabetically sorting the titles when you first grab them. That will provide a better long term user experience, since the fields will always be in the same order.

Mrod1598: Updated

jmwilliams89 (approved):

andykellr (approved):

PR #1451, opened 2023-08-08
feat(monitoring): OTLP APM

antonblock:
### Proposed Change

- Add `metrics` configuration section which will have parameters for collecting/exporting APM metrics
- Add sole `otlp` configuration to `metrics`, which takes parameters needed for the `otlpmetricgrpc` exporter
- `metrics` package mainly copies the `tracer` package pattern
- For now, have the OpAMP API report the current number of connected agents and counter of messages received from agents
- We'll setup deployment in a separate PR

Relevant config snippet to test:

```yaml
metrics:
  type: otlp
  interval: 10s
  otlp:
    endpoint: 127.0.0.1:4317
    insecure: true
```

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


cli/factory.go
```diff
@@ -337,6 +344,24 @@ func (f *Factory) BuildTracer(_ context.Context) (tracer.Tracer, error) {
...
+func (f *Factory) BuildMetricsProvider(_ context.Context) (metrics.Provider, error) {
+	cfg := &f.cfg.Metrics
+	interval := cfg.Interval
+	if interval.Seconds() == 0 {
+		interval = 1 * time.Minute
+	}
+
+	switch cfg.Type {
+	case config.MetricsTypeOTLP:
+		return metrics.NewOTLP(&cfg.OTLP, interval, tracer.DefaultResource())
```
andykellr, line 357: If I understand how this works correctly, I think we should factor out the resource from BuildTracer and BuildMetricsProvider in the event that we decide to add additional resource attributes in the future.

rest/api.go
```diff
@@ -1341,6 +1345,11 @@ func RolloutStart(c *gin.Context, bindplane exposedserver.BindPlane) {
 
 	bindplane.Logger().Debug("startRollout")
 	name := c.Param("name")
+	h, _ := mp.Int64Histogram("rest/rolloutStart", metric.WithUnit("ms"))
+	start := time.Now().UTC()
+	defer func() {
+		h.Record(ctx, time.Since(start).Milliseconds(), metric.WithAttributes(attribute.String("name", name)))
```
andykellr, line 1351: Do you think adding the name here will create cardinality concerns?

antonblock: Yeah I intended to remove this but forgot, testing locally it didn't seem super useful over the existing latency logs

config/metrics.go
```diff
@@ -0,0 +1,64 @@
...
+	"errors"
+	"fmt"
+	"time"
+)
+
+const (
+	// MetricsTypeOTLP is the OTLP metrics type
+	MetricsTypeOTLP = "otlp"
+	// MetricsTypeNop is the metrics type that does nothing
+	MetricsTypeNop = ""
```
cpheps, line 27: Maybe have this be an actual value that we default to? Empty string seems easy to miss if you're looking at a config.

config/override.go
```diff
@@ -212,6 +212,12 @@ func DefaultOverrides() []*Override {
 		NewOverride("tracing.otlp.endpoint", "the endpoint to send tracing data to, if using OTLP", ""),
 		NewOverride("tracing.otlp.insecure", "whether to use insecure TLS for tracing", false),
 
+		// Metrics overrides
+		NewOverride("metrics.type", "the type of metrics to use. One of: otlp", ""),
```
cpheps, line 216: Can we use the constant of `MetricsTypeNoop` here so if we change that value it carries over?

internal/opamp/api.go
```diff
@@ -119,6 +120,8 @@ type opampServer struct {
 
 var _ protocol.Protocol = (*opampServer)(nil)
 var _ opamp.Callbacks = (*opampServer)(nil)
+var mp metric.Meter = otel.Meter("opamp/agents")
+var agentMessages int64
```
cpheps, line 125: Should this be an atomic? Seems like it could be modified while we observe it.

internal/opamp/legacy/api.go
```diff
@@ -199,6 +224,8 @@ func (s *legacyOpampServer) OnConnected(_ legacyOpamp.Connection) {
 	defer span.End()
 }
 
+var agentMessages int64
```
cpheps, line 227: Same here

internal/opamp/api.go
antonblock: I'm not too worried about that since +-1 isn't a big deal, but also doesn't seem like a big deal to add that

config/metrics.go
```diff
@@ -0,0 +1,64 @@
...
+	Interval time.Duration `mapstructure:"interval,omitempty" yaml:"interval,omitempty"`
+
+	// OTLP is the config for sending OTLP metrics
+	OTLP OTLPMetrics `mapstructure:"otlp,omitempty" yaml:"otlp,omitempty"`
+}
+
+// OTLPMetrics is the config for sending OTLP metrics
+type OTLPMetrics struct {
+	// Endpoint is the endpoint to send metrics to
+	Endpoint string `mapstructure:"endpoint,omitempty" yaml:"endpoint,omitempty"`
```
jsirianni, line 45: Is this the GRPC endpoint? I think we should note which protocol we are using.

config/config.go
```diff
@@ -105,6 +111,10 @@ func (c *Config) Validate() error {
 		return fmt.Errorf("failed to validate logging: %w", err)
 	}
 
+	if err := c.Metrics.Validate(); err != nil {
+		return fmt.Errorf("failed to validate metrics: %w", err)
+	}
```
jsirianni, line 116: No action needed here, but wanted to mention that we should consider that upgrades will not include a metric configuration unless the user runs init. I think the upgrade path is fine right now, but we should keep that in mind.

jsirianni (commented):
Connected agents metric is showing `0` for me, but I have one agent.

Screenshot from 2023-08-09 12-41-25

config/metrics.go
antonblock: This is missing from tracing, but added

config/metrics.go
```diff
@@ -0,0 +1,64 @@
...
+
+	// Insecure is whether to use an insecure connection
+	Insecure bool `mapstructure:"insecure,omitempty" yaml:"insecure,omitempty"`
+}
+
+// Validate validates the metrics config
+func (m *Metrics) Validate() error {
+	switch m.Type {
+	case MetricsTypeOTLP:
+		if m.OTLP.Endpoint == "" {
```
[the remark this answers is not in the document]
antonblock: Added

config/metrics.go
antonblock: I think I'm preferring the blank string because it means not putting in a `metrics` section will result in nothing happening and no logging, and it's consistent with our solution for tracing

config/metrics.go
jsirianni: Sorry I was not clear. I think `metrics.otlp.endpoint` is fine it was more of a documentation comment. 

```go
// Endpoint is the OTLP gRPC endpoint to send metrics to
Endpoint string `mapstructure:"endpoint,omitempty" yaml:"endpoint,omitempty"`
```

I think it is worth it to stay consistent with the trace config.

jsirianni (approved):

cpheps (approved):

andykellr (approved):

PR #1452, opened 2023-08-09
testing: Add reason the AgentUpdater function

shazlehu:
### Proposed Change

- Adds `Reason` to Event struct
- AgentUpdater returns a "reason" string that's stored in the Event to assist in reasoning about system behavior
- updateOrUpsertAgentTx checks if an update is a top, and doesn't notify in that case

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


shazlehu:
>

store/store.go
```diff
@@ -221,7 +221,7 @@ var ErrDoesNotSupportHistory = errors.New("store does not support resource histo
 // AgentUpdater is given the current Agent model (possibly empty except for ID) and should update the Agent directly. We
 // take this approach so that appropriate locking and/or transactions can be used for the operation as needed by the
 // Store implementation.
-type AgentUpdater func(current *model.Agent)
+type AgentUpdater func(current *model.Agent) (reason string)
```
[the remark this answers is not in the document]
andykellr: I worked on this with Sam and while there aren't currently examples of this, I think it's reasonable that at some point the update could depend on the agent that is being modified. In other words, the reason for the update may be different depending on the agent. Also, we may wish to include additional information in the reason, like "Upgrade v1.24.0 to v1.26.0". Without access to the agent, we wouldn't be able to include the existing version.

graphql/model/models.go
```diff
@@ -94,9 +94,10 @@ func ToGraphMetric(m *record.Metric) (*GraphMetric, error) {
...
 
 // ClearCurrentAgentUpgradeError clears the current agent upgrade error
-func ClearCurrentAgentUpgradeError(cur *model.Agent) {
+func ClearCurrentAgentUpgradeError(cur *model.Agent) string {
 	if cur.Upgrade == nil {
-		return
+		return "clear empty upgrade error"
 	}
 	cur.Upgrade.Error = ""
+	return "clear upgrade error"
```
andykellr, line 102: Actually @BinaryFissionGames here is an example of a dynamic reason.

andykellr (approved):

andykellr (approved):
There's a test that needs to be fixed, likely because of agent updates now ignoring noops.

graphql/model/models.go
BinaryFissionGames: Yeah, I did see this case but I thought we would never see the first message (`clear empty upgrade error`) because it leaves the agent unchanged, so I didn't think they dynamic return reason was beneficial here.

andykellr: Ah, that's fair in this case. I still like the flexibility it allows, but I don't feel that strongly.

internal/opamp/legacy/updater.go
```diff
@@ -119,12 +122,68 @@ func (u *updater) handleUpdates(ctx context.Context, updates store.BasicEventUpd
...
+		finalHash := h.Sum(nil)
+
+		// hash the Configuration
+		// Convert agent.Configuration to a string, if necessary
+		configurationStr := fmt.Sprintf("%v", agent.Configuration)
+		h = sha256.New()
+		h.Write([]byte(configurationStr))
+		configurationHash := h.Sum(nil)
+
+		u.logger.Info("agent update", zap.String("agentID", agent.ID),
```
shazlehu, line 173: Do we want this logged out?  Or change it to debug?

store/store.go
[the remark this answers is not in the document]
BinaryFissionGames: Sounds like we prefer the flexibility, which makes sense. 

@shazlehu Could we update the comment above this type definition explaining what reason is (that it's used for debugging)?

internal/opamp/legacy/updater.go
BinaryFissionGames: IMO we should change this to debug, seems like it'd be very noisy.

PR #1459, opened 2023-08-11
fix: CLI tables should be sorted by the first column

andykellr:
### Proposed Change
The CLI does not always print tables with sorted rows. The should be sorted by the values in the first column.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


dsvanlani (approved):

cli/printer/table.go
```diff
@@ -75,3 +81,13 @@ func (tp *TablePrinter) Reset() {
 	tp.table.ClearFooter()
 	tp.table.SetHeader([]string{})
 }
+
+type byFirstItem [][]string
+
+func (a byFirstItem) Len() int           { return len(a) }
+func (a byFirstItem) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
+func (a byFirstItem) Less(i, j int) bool { return strings.ToLower(a[i][0]) < strings.ToLower(a[j][0]) }
```
BinaryFissionGames, line 89: Is there any chance that `len(a[n]) == 0` ever? Seems like it's fair to make the assumption it won't, just asking because this could possible be index out of bounds.

cli/printer/table_test.go
```diff
@@ -0,0 +1,154 @@
...
+}
+func (tp *testPrintable) PrintableKindPlural() string {
+	return "tests"
+}
+func (tp *testPrintable) PrintableFieldTitles() []string {
+	return tp.titles
+}
+func (tp *testPrintable) PrintableFieldValue(title string) string {
+	return tp.values[title]
+}
```
BinaryFissionGames, line 46: nit: Any reason for a test struct over a generated mock?

cli/printer/table.go
andykellr: I was going to check for it and then make a test case but it's not realistic that we would create a model with no printable titles.

cli/printer/table_test.go
andykellr: I think it's easier to read in simple cases like this where you would need several mocks with different mocked results for values.

BinaryFissionGames (approved):

PR #1460, opened 2023-08-11
fix: Set edge ActiveTypeFlag based on the intersection of source and target nodes

dsvanlani:
### Proposed Change
We were seeing issues that edges between the source and destination processor were not reflecting the `activeTypeFlag` of the pipeline they were in, first tried to fix in https://github.com/observIQ/bindplane-op/pull/1398.  This introduced the same bug though, but from the other side.

We now set the `attributes[AttributeName.ActiveTypeFlag]` value for edges based on the *intersection* of the active type flags of the source and target nodes.

i.e.  Previously we see an animated line between the file source processors and logging destination processors
image

Now we correctly see:
image

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


shazlehu (approved):
Looks good, works well

ui/src/utils/graph/utils.ts
```diff
@@ -396,12 +399,31 @@ export function getNodesAndEdges(
...
+      // Get the ActiveTypeFlag attribute from the source and target nodes
+      const sourceNodeActiveTypeFlag =
+        sourceNode.data.attributes[AttributeName.ActiveTypeFlags];
+      const targetNodeActiveTypeFlag =
+        targetNode.data.attributes[AttributeName.ActiveTypeFlags];
+
+      // If the source and target nodes have ActiveTypeFlag attributes
+      if (sourceNodeActiveTypeFlag && targetNodeActiveTypeFlag) {
+        // Set the ActiveTypeFlag attribute as the intersection of the two nodes
+        edge.data.attributes = {
```
antonblock, line 419: I think we also need to set `data.attributes: {}` on line 396-397. Otherwise the page bombs out

```
Uncaught TypeError: Cannot read properties of undefined (reading 'activeTypeFlags')
    at isNodeDisabled (nodeUtils.ts:22:1)
    at getNodesAndEdges (bundle.js:40751:98)
```

antonblock (approved):

PR #1461, opened 2023-08-14
fix(UI): Check existing names in handleSubmit

antonblock:
### Proposed Change

- Pass the existing resource names to `initFormErrors` in the handler for the save button
- This will set the error at the form level and prevent the request from being made

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


Mrod1598 (approved):
Doesn't let me save a bad destination anymore

shazlehu (approved):

PR #1493, opened 2023-08-22
feat: Timeout server requests after 20 seconds

dsvanlani:
### Proposed Change
This PR ensures that 1) requests to the server via GraphQL or REST API timeout after 20 seconds, and 2) the context being passed to store methods is canceled after 20 seconds.

Changes:
- Adds a server timeout middleware to authV1 that returns a 503 after 20 seconds and cancels the `request.Context()` for the gin context.
    - This does not apply to websocket connections
    - The 5xx status is necessary for GraphQL client to not keep trying the request.
-  Make sure that all REST API methods are passing the `request.Context` into the store methods they use.
- Update the write/read timeout for the server to 2 mins, this is consistent with Enterprise.
- Removes the `SetTimeout` of the bindplane client, this is handled in the server now.

#### REST Timeout

Easily tested by adding sleeps to rest and graphql methods.  E.g in rest/api.go#Agents:

```go
func Agents(c *gin.Context, bindplane exposedserver.BindPlane) {
	ctx, span := tracer.Start(c.Request.Context(), "rest/Agents")
	defer span.End()

	time.Sleep(21 * time.Second)

```

Using the CLI:

```
> bindplane get agents
Error: unable to get agents, got 503 Service Unavailable
exit status 1
```

Using Curl

```
> curl -u admin:admin http://localhost:3001/v1/agents
{"message":"Request timed out"}%                                                              
```

#### GraphQL Timeout

On the Playground:
image

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


middleware/server_timeout.go
```diff
@@ -0,0 +1,52 @@
...
+	"context"
+	"net/http"
+	"time"
+
+	"github.com/gin-contrib/timeout"
+	"github.com/gin-gonic/gin"
+)
+
+// DEFAULT_TIMEOUT_DURATION is 20 seconds.
+const DEFAULT_TIMEOUT_DURATION = 20 * time.Second
```
cpheps, line 13: nit: This const should be named `DefaultTimeoutDuration` all variables in Go should be Camel case unless they are representing an abbreviation or acronym.

cpheps (changes requested):

cli/commands/serve/server.go
```diff
@@ -42,10 +42,10 @@ import (
...
 
 const (
-	defaultReadTimeout     = 20 * time.Second
-	defaultWriteTimeout    = 20 * time.Second
-	defaultIdleTimeout     = 60 * time.Second
-	defaultShutdownTimeout = 20 * time.Second
+	defaultReadTimeout     = 2 * time.Minute
+	defaultWriteTimeout    = 2 * time.Minute
+	defaultIdleTimeout     = 2 * time.Minute
+	defaultShutdownTimeout = 2 * time.Minute
```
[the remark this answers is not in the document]
cpheps: Yeah, this is outside of this PR but maybe we just should expose these to just use the same values in enterprise.

middleware/server_timeout.go
```diff
@@ -0,0 +1,52 @@
...
+// a timeout to the request context.
+func timeoutHandler(timeoutDuration time.Duration) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		// Don't worry about canceling context for websocket connections
+		if c.IsWebsocket() {
+			c.Next()
+			return
+		}
+
+		ctx, cancel := context.WithTimeout(c.Request.Context(), timeoutDuration)
```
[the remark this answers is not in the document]
cpheps: After talking with Dave it sounds like `c.Next()` actually blocks so this should be fine.

middleware/server_timeout.go
```diff
@@ -0,0 +1,82 @@
...
+		timeout.WithTimeout(timeoutDuration),
+		timeout.WithResponse(timeoutResponse),
+	)
+}
+
+// TimeoutMiddleware returns a gin.HandlerFunc (middleware) that adds
+// a timeout to the request context of 20 seconds.
+func TimeoutMiddleware() gin.HandlerFunc {
+	return timeoutMiddlewareWithDuration(defaultTimeoutDuration)
+}
```
BinaryFissionGames, line 82: We should pass in the duration here, in EE we are gonna need to bump up the timeout duration for the endpoint uploading the agent upgrade packages, which can take a long time (like, upwards of a minute).

cpheps: I had him remove it but I didn't think about putting it in EE. Sorry @dsvanlani

PR #1512, opened 2023-08-28
feat: Add store.UpdateAgentStatus

andykellr:
### Proposed Change
Adds store.UpdateAgentStatus which can be implemented more efficiently that using UpdateAgent.

Includes the following changes:
* Change from https://github.com/observIQ/bindplane-op/pull/1452 to ignore no-op calls to UpdateAgent.
* eventRecorder which will record update events to ensure that meet expectations in tests

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


store/event_recorder_test.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+func (r *eventRecorder) record(ctx context.Context, c <-chan BasicEventUpdates) {
+	go func() {
+		for {
+			select {
+			case update := <-c:
+				r.mtx.Lock()
+				r.events = append(r.events, update)
+				r.mtx.Unlock()
```
cpheps, line 62: Is events being read at the same time it's being written? This lock seems unnecessary if we're writing then reading.

store/store.go
```diff
@@ -72,6 +72,13 @@ type Store interface {
 	// agent will be nil. An error is only returned if the update fails.
 	UpdateAgent(ctx context.Context, agentID string, updater AgentUpdater) (*model.Agent, error)
 
+	// UpdateAgentStatus will update the status of an existing agent. If the agentID does not exist, this does nothing. An
+	// error is only returned if updating the status of the agent fails.
+	//
+	// In some store implementations this will be more efficient than using UpdateAgent and modifying the Agent model
+	// directly.
```
cpheps, line 79: Questioning if this comment is needed since we're not making a distinction on when to use this based on the underlying implementation of the interface.

cpheps (approved):

shazlehu (approved):

store/store.go
andykellr: > Questioning if this comment is needed since we're not making a distinction on when to use this based on the underlying implementation of the interface.

Is this better?

```
When only the agent status needs to be modified, this should be preferred over UpdateAgent. In some store implementations this will be more efficient.
```

store/event_recorder_test.go
andykellr: It is read in assertEvents inside of the `assert.Eventually` which also locks.

PR #1513, opened 2023-08-28
feat: Make Measurements Interval Configurable

Mrod1598:
### Proposed Change

- make measurements collection interval configurable on a per config basis

#### Testing

1. Create a config in the ui that will collect some data
2. use the cli to export it to a file using the `--export` flag 
3. Update the config with a new measurement interval:
Ex:
```
spec:
    contentType: ""
    measurementInterval: 60s
```

4. Reapply the config and roll it out
5. watch the server to make sure the collection are coming in at the right interval 

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


model/configuration.go
```diff
@@ -1377,7 +1388,7 @@ func (c *Configuration) determinePipelineTypeUsage(ctx context.Context, store Re
 
 	// the agent ID, URL, and tls values aren't important
 	rc := &renderContext{
-		RenderContext:     otel.NewRenderContext("AGENT_ID", c.Name(), "BINDPLANE_URL", false, nil),
+		RenderContext:     otel.NewRenderContext("AGENT_ID", c.Name(), "BINDPLANE_URL", false, nil, "10s"),
```
cpheps, line 1391: Is this suppose to be hard coded to `10s`

Mrod1598: Nope. must have missed that when messing with tests....

model/otel/configuration.go
```diff
@@ -434,14 +436,18 @@ func (c *Configuration) AddAgentMetricsPipeline(rc *RenderContext, headers map[s
...
+
 	parts := Partial{
 		Receivers: ComponentList{{
 			"prometheus/_agent_metrics": map[string]any{
 				"config": map[string]any{
 					"scrape_configs": []map[string]any{
 						{
 							"job_name":        "observiq-otel-collector",
-							"scrape_interval": "10s",
+							"scrape_interval": rc.measurementInterval,
```
BinaryFissionGames, line 450: Is there a reason we pass this on the render context instead of accessing it directly? (c.Spec.MeasurementInterval)

Mrod1598: because there's two different models. this is within the otel package so we don't have access to the struct anymore.

BinaryFissionGames: Ahh, I see now, totally missed that.

BinaryFissionGames (approved):

cpheps (approved):

PR #1514, opened 2023-08-28
fix: Rollout Update Batching

cpheps:
### Proposed Change
- Removed rollout update eventbus and added a local RolloutBatcher
  - The batcher will batch rollout updates from agents over 100ms then call `UpdateRollout` once per config per batch

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


store/rollout_batcher.go
```diff
@@ -0,0 +1,175 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			d.logger.Error("Context error while batching events", zap.Error(ctx.Err()))
+			return ctx.Err()
+		case d.eventChan <- configStatuses:
+			return nil
+		}
+	}
+}
```
BinaryFissionGames, line 118: I think this bottom for loop doesn't need to exist (every path exits, never loops)

cpheps: Good call I rearranged things and left it in

BinaryFissionGames (approved):
Seems to be working good for me!

store/boltstore.go
```diff
@@ -57,12 +57,19 @@ func NewBoltStore(ctx context.Context, db *bbolt.DB, options Options, logger *za
...
 		BoltstoreCore: &BoltstoreCore{
 			DB:             db,
-			StoreUpdates:   NewUpdates(ctx, options, logger, BuildBasicEventBroadcast(), BuildRolloutEventBroadcast(), NewRolloutUpdates),
 			Logger:         logger,
+			RolloutBatcher: NewNopRolloutBatcher(),
 			SessionStorage: NewBPCookieStore(options.SessionsSecret),
 		},
 	}
 
+	// There is a cyclic dependency here that's not great where the rollout batcher needs the store and the updates need the rollout batcher.
```
andykellr, line 66: Since it only depends on one function, you could create a local interface like this. It simplifies testing as well.

```go
// UpdateRollout implements a method to update a rollout for a given configuration
//
//go:generate mockery --name=UpdateRollout --inpackage --with-expecter --filename=mock_update_rollout.go --structname=MockUpdateRollout
type UpdateRollout interface {
	UpdateRollout(ctx context.Context, configurationName string) (*model.Configuration, error)
}
```

cpheps: This is true but the cyclic dependency still exists. You need a store for the batcher and you need a batcher for the EventUpdates. But you can't create a store without event Updates. 

Really the EventUpdates just needs a batcher but that can't be isolated currently outside of the Storage constructor.

cpheps: Unless I'm missing what your proposing since the Store would still be the thing to implement the `UpdateRollout` interface.

andykellr: No, it's still cyclic. It would just avoid having another full mock store in tests. No big deal.

andykellr (approved):

PR #1521, opened 2023-08-31
feat: Update index interface to use context

jmwilliams89:
### Proposed Change
- Update index interface to use context
- Add tracing to graphql resolver
- Update mocks

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


store/mapstore.go
```diff
@@ -790,7 +790,7 @@ func (mapstore *mapStore) upsertAgent(agentID string, updater AgentUpdater, upda
 	mapstore.agents[agentID] = agent
 
 	// update the index
-	err := mapstore.agentIndex.Upsert(agent)
+	err := mapstore.agentIndex.Upsert(context.Background(), agent)
```
cpheps, line 793: Should there be a hole punched through upsertAgent to get the context here or is that a different PR?

cpheps (approved):

jsirianni (approved):

store/mapstore.go
jmwilliams89: I avoided it here since mapstore should be deprecated. It didn't feel worth the effort.

PR #1539, opened 2023-09-07
chore: Refactor agent's indexing to allow for efficient rollout indexing

BinaryFissionGames:
### Proposed Change
* break the agent's indexing of rollout status into it's own indexer object so that indexing only rollout fields can be done more efficiently, while also maintaining the same logic as the agent indexing.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


model/agent.go
```diff
@@ -693,3 +668,68 @@ var AgentFieldAccessor FieldAccessor[*Agent] = func(field string, item *Agent) s
...
+	FieldRolloutError:         {},
+	FieldRolloutPending:       {},
+	FieldRolloutWaiting:       {},
+	FieldRolloutComplete:      {},
+}
+
+// IndexFields calls index with the key-value pairs that should be indexed
+func (a AgentRolloutStatusIndexer) IndexFields(index modelSearch.Indexer) {
+	// index the current version
+	index(FieldConfigurationCurrent, a.ConfigurationStatus.Current)
```
BinaryFissionGames, line 703: Technically, I don't think we'll need FieldRolloutComplete or FieldRolloutComplete for the downstream usage, but it's hard to justify leaving them out here.

model/agent.go
```diff
@@ -693,3 +668,68 @@ var AgentFieldAccessor FieldAccessor[*Agent] = func(field string, item *Agent) s
...
+// AgentRolloutStatusIndexerFieldSet is a set of fields returned by
+// AgentRolloutStatusIndexer.IndexFields.
+var AgentRolloutStatusIndexerFieldSet = map[string]struct{}{
+	FieldConfigurationCurrent: {},
+	FieldConfigurationPending: {},
+	FieldConfigurationFuture:  {},
+	FieldRolloutError:         {},
+	FieldRolloutPending:       {},
+	FieldRolloutWaiting:       {},
+	FieldRolloutComplete:      {},
```
andykellr, line 697: I think `"configuration"` should be here since it is changed below and I think we should move `"status"` in here as well since the other use of this code will be in `UpdateAgentStatus`.

BinaryFissionGames: Good catch. Moved status down here and made sure configuration and status were in this set.

model/agent.go
```diff
@@ -213,24 +234,7 @@ func (a *Agent) UniqueKey() string {
...
-	case Deleted:
-		return "Deleted"
-	case Configuring:
-		return "Configuring"
-	case Upgrading:
-		return "Upgrading"
-	default:
-		return "Unknown"
-	}
+	return a.Status.String()
```
andykellr, line 237: I think this change is fine. The only reason I didn't do this originally is that I didn't want to suggest that these string values were just the names of the values, e.g. ComponentFailed becomes "Component Failed" and we add "Unknown" as a possibly result.

I might prefer to use a function `DisplayText()` (which feels editorial to me) instead of `String()` (which feels like it might be a default implementation), but I don't feel strongly.

BinaryFissionGames: Yeah, the reason I used `String()` here is so that it implemented the `Stringer` interface, which is used in all sorts of fun places that could be useful to us.

For instance, zap has a stringer field type: https://pkg.go.dev/go.uber.org/zap#Stringer
and using the printf family of functions will use the string value from a Stringer (https://go.dev/play/p/ZU05OmSojan) - I thought that could be useful for us.

I've changed it to display text, since it seems we maybe don't want that behavior?

andykellr (approved):

shazlehu (approved):

PR #1540, opened 2023-09-07
Fix/escape hash in eval

dsvanlani:
A customer reported an issue of being unable to save a config when using a value with a `#` character in the filter fields processor.

Changes:
Put all parenthetical expressions into a scalar to escape any `#` character.

### Testing

You should be able to use `#` in all fields in the filter_fields processor.  E.g.

image


model/testfiles/destinationtype-cabin.yaml
```diff
@@ -18,7 +18,6 @@ spec:
       required: true
   logs:
     processors: |
-      # batch should be last
```
andykellr, line 21: I'm guessing this removes the comment because replacing it would break the yaml. I think our solution should still allow for comments. These things occasionally get complicated.

dsvanlani: Well, comments are still fine but just not in the templated otel configuration.  We've never needed this functionality or elected to use it before now, and we can still comment in our resource-types.  Do you feel strongly about this?

andykellr (approved):

BinaryFissionGames (approved):

PR #1560, opened 2023-09-11
deps: Use Node 18

antonblock:
##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


BinaryFissionGames:
If you're using nodenv to manage node versions, you might need to upgrade to be able to install v18.17.1: `brew upgrade nodenv node-build`

dsvanlani (approved):

PR #1574, opened 2023-09-12
feat: Add flag to control metric suffixes for prometheus destinations

BinaryFissionGames:
### Proposed Change
* Adds a flag that was recently introduced to all prometheus destinations.

This option will only take effect in v1.34.0 of our agent, but the flag has been present since v1.32.0.
The default for the underlying exporter is actually `true`, so we need to default to false to avoid breaking metric name changes for v1.34.0+ agents.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


antonblock (approved):

cpheps (approved):

jsirianni (approved):
Looks good. Tested against 1.33 and 1.34 pre release agent versions.

PR #1575, opened 2023-09-12
fix: AgentsPerPhase overflow

shazlehu:
### Proposed Change

AgentsPerPhase was overflowing, creating a negative value. This change eliminates the use of Math.Pow, and only does the number of multiplications necessary to determine if we've hit the maximum value.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


model/configuration.go
```diff
@@ -324,10 +323,22 @@ type RolloutProgress struct {
 
 // AgentsPerPhase returns the number of agents that will be updated in the current phase.
 func (r *Rollout) AgentsPerPhase() int {
-	numAgents := r.Options.PhaseAgentCount.Initial * int(math.Pow(r.Options.PhaseAgentCount.Multiplier, float64(r.Phase)))
-	if numAgents > r.Options.PhaseAgentCount.Maximum {
-		return r.Options.PhaseAgentCount.Maximum
+	multiplier := r.Options.PhaseAgentCount.Multiplier
```
cpheps, line 326: Is there anyway to store state or skip this call after we reach the maximum? Once we hit the maximum we should always just use that right so we don't need to recalculate this.

I don't think it's a big deal as this function is pretty light weight more curious if we have the ability to detect that so we don't have to recalculate everytime.

cpheps (approved):

model/configuration.go
```diff
@@ -324,10 +323,22 @@ type RolloutProgress struct {
...
+	// If the multiplier is 1 or less, the number of agents won't change over phases.
+	if multiplier <= 1 {
+		return r.Options.PhaseAgentCount.Initial
 	}
+
+	numAgents := r.Options.PhaseAgentCount.Initial
+	for i := 0; i < r.Phase; i++ {
+		numAgents *= int(multiplier)
```
BinaryFissionGames, line 334: I think we need to be careful here, multiplier doesn't need to be integral, it also supports multipliers like `1.5`, but here it's effectively treated like `1`.

I think the correct way is to calculate a cumulative multiplier as a float64, then multiply it by initial when returning or comparing.

BinaryFissionGames: Maybe something like:
```go
	totalMult := float64(1)
	initial := r.Options.PhaseAgentCount.Initial
	for i := 0; i < r.Phase; i++ {
		totalMult *= multiplier
		numAgents := int(float64(initial) * totalMult)

		// Check for potential overflow or if the number exceeds the max.
		if numAgents >= r.Options.PhaseAgentCount.Maximum || numAgents < 0 {
			return r.Options.PhaseAgentCount.Maximum
		}

	}

	return int(float64(initial) * totalMult)
```

shazlehu: You're right, we should support float values for the multiplier as expected

BinaryFissionGames (approved):

PR #1577, opened 2023-09-14
fix(measurements): Change source IDs in pipeline

antonblock:
### Proposed Change

- Change source names to include ULID in addition to index

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


dsvanlani (approved):

model/configuration.go
```diff
@@ -1204,6 +1204,7 @@ func (c *Configuration) Graph(ctx context.Context, store ResourceStore) (*graph.
 
 	for i, source := range c.Spec.Sources {
 		sourceName := TrimVersion(source.localName(KindSource, i))
+		// sourceName := TrimVersion(fmt.Sprintf("%s%d_%s", strings.ToLower(string(KindSource)), i, source.ID))
```
BinaryFissionGames, line 1207: Seems like we're good with deleting this line?

model/configuration_test.go
BinaryFissionGames, line 1: Side note, these tests are ALWAYS a pain to update, would be cool to update these tests so that the "golden" configs are generatable files that we could just end up checking the diff on.

BinaryFissionGames (approved):
Working well for me, including the overview page.

PR #1584, opened 2023-09-15
feat: Add featuregate to resourceTypes

Mrod1598:
### Proposed Change

1. Add featuregate to resourceTypes
2. Set seeded resources to the base gates

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


model/resource_type.go
```diff
@@ -325,6 +328,20 @@ func (rt *ResourceType) evalTemplate(r ResourceTypeTemplate, nameProvider otel.C
...
+// ----------------------------------------------------------------------
+// featureGate
+// ----------------------------------------------------------------------
+
+// featureGate returns the feature flag for this resource type
+func (rt *ResourceType) FeatureGate() string {
+	return rt.Spec.FeatureGate
+}
+
+// IsFeatureEnabled returns true if the feature flag for this resource type is enabled
```
cpheps, line 340: Comment is wrong

model/resource_type.go
```diff
@@ -325,6 +328,20 @@ func (rt *ResourceType) evalTemplate(r ResourceTypeTemplate, nameProvider otel.C
 	return set
 }
 
+// ----------------------------------------------------------------------
+// featureGate
+// ----------------------------------------------------------------------
+
+// featureGate returns the feature flag for this resource type
```
cpheps, line 335: nit: Comment should be capitalized like function name

store/store.go
```diff
@@ -409,6 +409,20 @@ func GetSeedResources(ctx context.Context, files embed.FS, folders []string) ([]
...
+		switch rType := resource.(type) {
+		case *model.SourceType:
+			rType.SetFeatureGate("base-sources")
+		case *model.DestinationType:
+			rType.SetFeatureGate("base-destinations")
+		case *model.ProcessorType:
+			rType.SetFeatureGate("base-processors")
+		}
+		allEmbedded[resourceIndex] = resource
+	}
```
BinaryFissionGames, line 424: Forgot about this when we were looking over this together, but these should be "defaults" (only set if there is no feature gate).

Mrod1598: Good point, updated.

model/resource_type.go
Mrod1598: Updated.

model/resource_type.go
Mrod1598: Updated.

cpheps (approved):

BinaryFissionGames (approved):

PR #1585, opened 2023-09-18
Trim version from destination

shazlehu:
### Proposed Change

We were seeing multiple destinations in the overview page as the result of broken config. This change should mitigate that.
##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


dsvanlani (approved):

andykellr (approved):
That has to be my favorite map value for creating a set.

dsvanlani (approved):

dsvanlani:
> That has to be my favorite map value for creating a set.

Leaves no ambiguity!

PR #1586, opened 2023-09-18
chore: Refactor ApplySortOffsetAndLimit, AssignMetricsToGraph

dsvanlani:
### Proposed Change
These are changes necessary for the server-side paginated agents table for our enterprise addition.

**Changes**
- Separate the offset and limit implementation of `ApplySortOffsetAndLimit` into separate `ApplyOffsetAndLimit` function.  To be used in the graphql resolver in enterprise.
- `AssignMetricsToGraph` is exported and now returns an error when it fails to convert a metric into a graph metric.  It no longer needs to take a `bindplane.Server` instance because we don't use the logger.
- `AssignMetricsToGraph` assigns the agentID field of a graph metric.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


BinaryFissionGames (approved):
Working great for me 👍

Mrod1598 (approved):

PR #1603, opened 2023-09-20
feat: Azure Blob Storage Source

cpheps:
### Proposed Change
Added the Azure Blob Storage Source that exposes the [azureblob receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/azureblobreceiver)

### Source Configuration
Screenshot 2023-09-20 at 12 46 22 PM

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


resources/source-types/azure_blob.yaml
```diff
@@ -0,0 +1,91 @@
...
+      description: Name of the Azure Storage container where traces are stored.
+      type: string
+      required: false
+      default: traces
+      relevantIf:
+        - name: enable_traces
+          operator: equals
+          value: true
+
+  logs:
```
jsirianni, line 71: I think this can be combined into one pipeline configuration 
```suggestion
  logs+traces:
```
The order might matter ^

cpheps: It can be combined bit it also makes it so the toggles don't matter. I like if you toggle off traces it won't actually give you a trace exporter. If we combine them I think the best I can do is just not configure the trace container, which has a default value if omitted.

jsirianni: 🤦 you are correct on that

jsirianni (approved):

resources/source-types/azure_blob.yaml
```diff
@@ -0,0 +1,91 @@
...
+        - text: Read more
+          url: https://learn.microsoft.com/en-us/azure/storage/blobs/storage-blob-event-overview
+      type: string
+      required: true
+
+    - name: enable_logs
+      label: Enable Logs
+      description: Enable to collect logs.
+      type: bool
+      default: true
```
Mrod1598, line 41: I believe we're defaulting to false for these now right? at least for processors. guess we don't have standards for sources yet.

Mrod1598 (approved):

PR #1620, opened 2023-09-25
feat(measurements): Filter out measurements with previous naming scheme

antonblock:
### Proposed Change

- This is the easiest option, but may not be working the way we want. In that case we need to implement this at the store level

### Testing

1. Start BPOP with a version before 1.28.1
2. Install a collector older than 1.35.1
3. Create a config and push it down to the collector
4. Let it collect measurements for a while
5. Stop bindplane and checkout this branch and restart it
6. Rollout config changes
7. Observe that the graphql response for measurements includes only sources and processors with IDs and not any without.
8. Extra confirmation to check measurements in DB to verify old measurements still exit. Note: They may be cleared out if not check fast enough.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


Mrod1598 (approved):

Mrod1598:
Not seeing anything in gql that doesn't have an ID

cpheps (approved):
Verified metrics are in the database but not being shown in the UI.

